PR #253, opened 2016-11-13
Updated itemget and itemwield

Psionics-ToEE:
I am unsure if all of this is correct. If it is however, I have concerns about itemget. If it can insert into invisible inventory slots when the visible one is full, all users of the function will need to be cautious of using it or else they might make items disappear when they wanted them to go to the visible inventory!

### DIFF
--- a/TemplePlus/inventory.cpp
+++ b/TemplePlus/inventory.cpp
@@ -391,6 +391,28 @@ bool InventorySystem::IsRangedWeapon(objHndl weapon){
 	return false;
 }
 
+bool InventorySystem::IsVisibleInventoryFull(objHndl objHandle)
+{
+	auto obj = objSystem->GetObject(objHandle);
+	auto invenNumField = obj_f_critter_inventory_num;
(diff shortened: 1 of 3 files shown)


Psionics-ToEE:
It says checks have failed, I'm not sure what that means either.

Psionics-ToEE:
It's set to automatically put items in the invisible slot if the inventory is already full though. Which means that if a user is calling item_get under assumptions that the inventory has room, they will get the wrong outcome.

DudeMcDude:
Nevertheless the behavior of item_get up until now was such that it would not put the item in the inventory at all if it is full (and also return a 0 as a result). You should not change something like that because it may introduce bugs in scripts that did not anticipate this new behavior. Furthermore putting an item in an invisible slot means the user has no control over it, nor knowledge that it's even there.

Instead, I've added an optional flags input for item_get so if someone wants that behavior (putting items in invisible slots) they have to explicitly state that in the script:

    obj.item_get(item, IIF_Use_Max_Idx_200)

PR #254, opened 2016-11-22
Fixed vector access for prereq reads

Psionics-ToEE:
Tested before and after modification. Before, any feat with prereq values (or even attempting to write in there manually) crashed the code. After, game fully reaches main menu and plays without error.

### DIFF
--- a/TemplePlus/feat.cpp
+++ b/TemplePlus/feat.cpp
@@ -217,7 +217,6 @@ void LegacyFeatSystem::_GetNewFeatsFromFile()
 		auto featFile = tio_fopen(fmt::format("rules\\feats\\{}",f.name).c_str(), "rt");
 
 		char lineContent[2000]={0,};
-		FeatPrereq featPrereq;
 		NewFeatSpec featSpec;
 		while (tio_fgets(lineContent, 1000, featFile)){
 			auto len = strlen(lineContent);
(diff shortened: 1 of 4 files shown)


TemplePlus/feat.cpp
```diff
@@ -259,10 +260,26 @@ void LegacyFeatSystem::_GetNewFeatsFromFile()
 					
 					if (prereqArgCount < prereqCount)
 					{
-						featSpec.prereqs[prereqArgCount++].featPrereqCodeArg = atol(ch);
+						if (featSpec.prereqs.capacity() <= prereqArgCount)
```
shartte: Are you sure you wanted to use capacity() and not size() here? size() is the real current size of the vector.

TemplePlus/feat.cpp
```diff
@@ -259,10 +260,26 @@ void LegacyFeatSystem::_GetNewFeatsFromFile()
...
+							featSpec.prereqs.push_back(featPrereq);
+							prereqArgCount++;
+						}
+						else
+							featSpec.prereqs[prereqArgCount++].featPrereqCodeArg = atol(ch);
 					}
 					else
-						featSpec.prereqs[prereqCount++].featPrereqCode = atol(ch);
+					{
+						if (featSpec.prereqs.capacity() <= prereqCount)
```
shartte: Same question about capacity() vs. size() here

Psionics-ToEE: You're right, size was the correct value I wanted. I just made a commit to this pull, removed the featSpec allocation at the start of the function, and used size/resize to cut down code to just 4 lines.

Psionics-ToEE:
I also changed the unwield function in python layer to be able to have an optional argument to drop item instead of putting in inventory. Tested ingame and it works.

TemplePlus/python/python_object.cpp
```diff
@@ -2443,6 +2444,11 @@ static PyObject* PyObjHandle_Unwield(PyObject* obj, PyObject* args) {
 		es = EquipSlot::Invalid;
 	auto equipSlot = (EquipSlot)es;
 
+	if (item != objHndl::null && inventory.GetParent(item) == self->handle ) {
```
DudeMcDude: From what I see, the logic you propose is this:
If the optional arg used and is an item handle of **_any item that is in the wielder's inventory_**, then the unwield action should drop the item worn at slot **es**.

I don't understand the logic behind this, and it seems very "ad-hoc". It would be more intuitive for it to accept a flag (or integer) arg.

Psionics-ToEE: Yes, simply the usage of the optional argument is what triggers it. I figure if you don't want to drop an item, you don't put an item there, and if you do, the item handle is the minimum information required to call drop. Should I instead check that the item is first in an equip slot instead of just the inventory?

DudeMcDude: item_unwield currently operates on a specific item slot. Passing in the item handle is redundant in this sense, simply adding an integer arg should be enough and IMO less confusing.

Psionics-ToEE:
Changed the call to an optional integer flag and added a new overload for drop to use it with. Tested in game and it drops, didn't get any build errors or anything so unsure why I've got the red x above.

PR #266, opened 2017-12-03
Add pdb to TemplePlusSymbols-version.zip

### LINKED ISSUE #263: Can you post the .pdb for v1.0.56?
I have a .dmp file from a stack overflow crash I'm trying to debug, but the zip for the v1.0.56 release doesn't have the .pdb file required to open the .dmp file in vs2015. Unfortunately, I can't repro the bug; otherwise, I'd repro it against my own .exe and .pdb. Can you post the v1.0.56 .pdb from your build machine?

I built a release .exe and .pdb against the v1.0.56 tag, but the exe is a slightly different size, so the .exe and .pdb are incompatible with the .dmp file. Debugging in windows is very finicky, so it doesn't seem like I can get anywhere without the .pdb file created by the release build from the exact same environment (same toolchain, etc).

lucianposton:
Fixes #263

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,7 +19,7 @@ after_build:
   
   # Package the symbol files
   - dump_syms.exe Release\TemplePlus.pdb > TemplePlus.sym
-  - 7z a TemplePlusSymbols-%APPVEYOR_BUILD_VERSION%.zip TemplePlus.sym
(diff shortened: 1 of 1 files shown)


shartte:
Checked the size of the resulting PDB, which is about 18MB zipped. I think that's acceptable.
I wonder if the TemplePlus.pdb is enough to debug though, I am not certain where the debug info for our static libraries really resides. But we'll see.

shartte:
Oh, and thanks for the PR! :)

PR #275, opened 2018-01-31
Made Turn Undead accessible to python and added extra rage feat

doug1234:
Made turn unread accessible to python and added a message if a condition does not exist when using extend existing.  Also added the extend rage feat.


TemplePlus/python/python_dispatcher.cpp
```diff
@@ -153,13 +153,19 @@ PYBIND11_EMBEDDED_MODULE(tpdp, m) {
...
 					}
 					condStr.numArgs = cond->numArgs;
 					condStr.condName = cond->condName;
 					condStr.Register();
+				} else {
+					std::stringstream s;
+					s << "Extend Existing Error:  Condition ";
+				    s << condName;
+				    s << " does not exist!";
+					logger->info(s.str().c_str());
```
shartte: Here you can use the string formatting that the logger offers, which would look like:

    logger->info("Extend Existing Error: Condition {} does not exist!", condName);

doug1234: OK thanks!  I'll make the change.

DudeMcDude:
This introduces a small bug: when multiclassing a character by taking a Cleric level, it will initially lack Turn Undead. It will only manifest after a map transition. 
(bug might affect Paladins and Blackguards too, haven't checked)

doug1234:
Thanks for the review.  I didn't test multi-class characters but I did test with paladins and blackguards without any issues.  I'll look into multi-class characters tomorrow night and fix any issues.

doug1234:
I looked into this a bit.  It seems to be due to obj_f_critter_alignment_choice returning 0 instead of the value it should return.  Any thoughts on why this is happening?

DudeMcDude:
Yeah, because it changes the alignment_choice *after* applying the levelup (which, internally, calls the D20StatusRefresh).

I guess as a patch we can add another call to D20StatusRefresh inside uiCharEditor.FinishBtnMsg.

doug1234:
OK sounds good.  I'll let you know if that works.

PR #276, opened 2018-02-14
Extra smiting changes and updates to support the new feat.

doug1234:
Various changes to add the extra smiting feat.


TemplePlus/python/python_object.cpp
```diff
@@ -48,6 +49,10 @@
 
 namespace py = pybind11;
 
+// Helper function for checking if information should be updated with chargen packet info
+bool UpdateWithChargenPacketInfo(objHndl handle) {
```
DudeMcDude: Should add a comment on why this assumes IsSelectingFeats

doug1234: Thanks for the review.  I added the comment.  I'll try not to use hungarian in the future but I've been doing it for a long time...

PR #277, opened 2018-02-17
Added feat Extra Music

IPDTFE:
Unsure on how to modify the maximum on the radial menu, works normally nonetheless.

### DIFF
--- a/tpdatasrc/tpgamefiles/scr/feats/feat - Craven.py
+++ b/tpdatasrc/tpgamefiles/scr/feats/feat - Craven.py
@@ -6,4 +6,6 @@ def CheckPrereq(attachee, classLevelled, abilityScoreRaised):
 	#Sneak Attack Check
 	if not attachee.has_feat(feat_sneak_attack):
 		return 0
+	if attachee.stat_level_get(stat_level_paladin) > 1: #workaround until I figure out how to check for immunity to fear
+		return 0
 	return 1
--- a/tpdatasrc/tpgamefiles/scr/feats/feat - Craven.py
+++ b/tpdatasrc/tpgamefiles/scr/feats/feat - Craven.py
@@ -1,0 +1,9 @@
+from toee import *
+
+
+def CheckPrereq(attachee, classLevelled, abilityScoreRaised):
+	
+	#Sneak Attack Check
+	if not attachee.has_feat(feat_sneak_attack):
+		return 0
+	return 1
--- a/tpdatasrc/tpgamefiles/scr/tpModifiers/extra_music.py
+++ b/tpdatasrc/tpgamefiles/scr/tpModifiers/extra_music.py
@@ -1,0 +1,23 @@
+#Extra Music: SRD
+
+from templeplus.pymod import PythonModifier
+from toee import *
+import tpdp
+
+print "Registering Extra Music"
+
+def EMNewDay(attachee, args, evt_obj):
+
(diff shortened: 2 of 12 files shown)


doug1234:
For what its worth, it looks good to me.

IPDTFE:
Added [Vexing Flanker](http://dnd.arkalseif.info/feats/players-handbook-ii--80/vexing-flanker--3062/index.html)

doug1234:
FYI:  Vexing Flanker needs to be added to the fighter bonus feats list.

IPDTFE:
[I believe it already is?](https://imgur.com/a/eVDz3) Is the version I uploaded different?

IPDTFE:
Removed useless target line, it was a leftover of my previous coding attempt.

IPDTFE:
Added [Deft Opportunist](http://dnd.arkalseif.info/feats/complete-adventurer--54/deft-opportunist--572/index.html)

doug1234:
I might be wrong on this but don't you have to update the list of bonus feats in class011_fighter.py for it to register as a fighter bonus feat?

IPDTFE:
Added it to the bonus fighter feats list, just to be on the safe side

IPDTFE:
Fixed feat prerequisite, it asked for combat expertise rather than combat reflexes

IPDTFE:
Added [Craven](http://dnd.arkalseif.info/feats/champions-of-ruin--27/craven--484/index.html) and updated Condition.cpp to implement extra sneak attack damage.

IPDTFE:
it's a habit I took when I noticed that feats that extended basic character abilities were called even when they weren't present on the character, or so it seemed. Extra music returns the exact value nonetheless, since if you don't possess the feat it returns exactly the number of bardic music uses you'd have per day.

It's meant as a secondary check in case of an unforeseen call to the function although these checks can be removed if deemed useless and unoptimized.

tpdatasrc/tpgamefiles/scr/feats/feat - Craven.py
```diff
@@ -6,4 +6,6 @@ def CheckPrereq(attachee, classLevelled, abilityScoreRaised):
 	#Sneak Attack Check
 	if not attachee.has_feat(feat_sneak_attack):
 		return 0
+	if attachee.stat_level_get(stat_level_paladin) > 1: #workaround until I figure out how to check for immunity to fear
```
DudeMcDude: There's a dispatch for immunity checks. I think it's not yet exposed to Python, but you may wish to extend it.

IPDTFE:
Yeah, it's both present as a flag and in the python list currently.

tpdatasrc/tpgamefiles/scr/tpModifiers/extra_music.py
```diff
@@ -0,0 +1,23 @@
...
+
+print "Registering Extra Music"
+
+def EMNewDay(attachee, args, evt_obj):
+
+	BardicMusicCount = attachee.has_feat("Extra Music")
+
+
+	#Extra Music grants 4 additional uses of Bardic Music each time the feat is taken
+  #Max music per day remains 3, can't find a way to alter it. Works properly nonetheless.
```
DudeMcDude: If you mean the max on the radial menu, currently it just takes bard level (condition.cpp BardicMusicRadial). I suppose you can change it to max(bardLvl, arg0). You can also generalize it to take the value from a query if you anticipate modification by other feats.

PR #279, opened 2018-03-01
Added Divine Armor Feat

doug1234:


doug1234:
Were you referring to divine_spell_level_can_cast?  I don't think I tried that.  I'll see if it work tomorrow.

DudeMcDude:
My mistake, it checks what spell level you can cast. It was required for some prestige classes like Mystic Theurge.

But it should be fairly easy to create GetDivineCasterLevel by iterating all classes and checking IsDivineCastingClass.

doug1234:
Future proofing is a good argument.  I added the properties to the python objects.

doug1234:
I went ahead and added divine vigor also since I had it ready to go.

doug1234:
The divine vigor bonus now has a string.

doug1234:
Added the divine shield feat.

PR #284, opened 2018-04-11
Telling Blow and Favored Power Attack Feats

doug1234:


doug1234:
OK cool.  You made the change for me.  Have you tested the new way much?  I see some possible difficulty but I haven't have time to test yet.

DudeMcDude:
I haven't tested extensively since this was mostly meant as a hotfix for broken character classes. But I think it replicates the functionality for the most part.

Also I'm reverting the change in d20_status. Are you going to add every feat you edit there? That's unreasonable. Remind me what the purpose was?

doug1234:
I'll let you know if I see anything when I have a chance to look it over.  

Adding the feat conditions in d20_status was necessary to get them to be visible in python.   Is there a better way to accomplish that?  If so, I can change it to work the way you want.

DudeMcDude:
What do you mean by "necessary to get them to be visible in python"? The has_feat method? That shouldn't be related.

Also, are you aware of initFeats inside d20_status? It iterates over all feats and adds their associated modifiers, so it should cover everything already without having to manually add them.

doug1234:
The has_feats() method works fine.  It's the conditions associated with the feats that were not showing up.  I'll take a look into it Wednesday and let you know precisely what the issue is.

DudeMcDude:
Where would they show up? In the Dispatcher? If so that's another issue, not related to Python. But it shouldn't happen either due to initFeats.

PR #291, opened 2018-05-17
Divine Spell Power, Quicken Turn Undead, Fast Wild Shape, bug 071 fix, Staggering Strike, Sudden Extend, Sudden Still, Sudden Silent

doug1234:


doug1234:
Added quicken Turn Undead.

doug1234:
I thought DivineSpellPowerEffect = PythonModifier("Divine Spell Power Effect", 4) gave me 4 args.  My plan was to have extras I could use later...

doug1234:
I changed the feat to use the same parameter as the effect to avoid any possible issue.  But wasn't the effect the one that needed the four parameters?  Also added fast wild shape.

DudeMcDude:
Yeah that's my mistake, I didn't notice it was a different modifier. You can change it back.

doug1234:
Added Staggering Strike.  

Unless you feel strongly about it I'd just leave it the way it is.  Having both with 4 is less confusing and I don't think there is much downside.

BTW:  Is there anything else you need me to change before you can add the changes to the branch?

doug1234:
Added sudden extend.

doug1234:
Good suggestion.  I am now using the object.

doug1234:
I was making it a query to be similar caster level modify.  But what you are asking for makes more sense.  There is no object that seems perfect so I'll make my own that is only metamagic data.

doug1234:
Hopefully it looks good now.  Also added sudden still and silent.

PR #294, opened 2018-05-30
Fix segfault in pathing algorithm

lucianposton:
Had a crash when path was null, so simply added a conditional to check whether
it is null before use. Believe it occurred because a target had no path
due to other npcs blocking the way.

### DIFF
--- a/TemplePlus/ai.cpp
+++ b/TemplePlus/ai.cpp
@@ -1589,36 +1589,39 @@ BOOL AiSystem::ImprovePosition(AiTactic* aiTac){
 		if (curNum > 0) {
 
 			auto path = curSeq->d20ActArray[curSeq->d20ActArrayNum - 1].path;
+			if (path) {
 
-			auto lowerBound = 0.0f;
-			auto upperBound = path->GetPathResultLength();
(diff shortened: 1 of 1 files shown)


lucianposton:
Much easier to view diff with https://github.com/GrognardsFromHell/TemplePlus/pull/294/files?w=1 fyi

shartte:
Do you know how it happens that it is null in the first place? The action code is quite complex and I don't know it enough. Are we masking an underlying issue by inserting this null check here?

lucianposton:
> Do you know how it happens that it is null in the first place?

No, all I know is my conjecture in the initial post. I'm even not sure what the consequences are to this workaround beyond avoiding the crash. Unfortunately, running a program in wine in a debugger behaves rather poorly, so I didn't dig that deep into the issue.

> Are we masking an underlying issue by inserting this null check here?

Probably

DudeMcDude:
I've done this a little differently, but thanks for the attempt nevertheless.

DudeMcDude:
Question: might you have been using Ready Vs. Approach when this happened?

PR #306, opened 2018-07-24
reckless charge feat, pain touch feats and fix for bug 184

doug1234:


doug1234:
AOOs should only be disabled for the final effect as far as I can tell.  It was working properly for me in testing but I'll take another look at it.  It will probably be Monday when I can give it a good second look.

DudeMcDude:
I was mainly thinking about the first issue (i.e. critters in the cloud who did make their save). Since it's working properly, then great!

PR #307, opened 2018-08-11
Added lingering song, fixed leve 20 bard inspire courage bug, augment healing, Wolverine's Rage and Versatile Unarmed Strike

doug1234:


IPDTFE:
I haven't read all of the code for lingering song, but it seems you've added the 3.0 version from song and silence. Wouldn't it be better to use the [3.5 version](http://dnd.arkalseif.info/feats/complete-adventurer--54/lingering-song--1776/index.html)?

doug1234:
Good find.  I'll update it to specify the 3.5 version.

If I understand it correctly, the feat works the same in both versions:  5 rounds doubled or 5 round and 5 more rounds is 10 rounds (1 minute).

IPDTFE:
Except for inspire confidence, but the implementation seems to be correct for the 3.5 version

doug1234:
I see... the 3.5 version effects a different set of bard abilities than the 3.0 version.  The 3.5 version was the version I implemented despite citing the wrong book.

doug1234:
Sorry about the typo.  :(  I will fix that shortly.  

I don't understand your comment about OnDamage.  The bonus for augment healing is to healing.

DudeMcDude:
The SpellHeal() function internally does a dispatch for event type 22, but as it turns out it's performed by the heal receiver and not the heal giver. It also doesn't broadcast the spell id. So I guess a new event type is warranted. But I think it should be named "Spell Healing Bonus" rather than Healing Bonus, since it assumes a spell id in the d20query?
If you want to broaden it to include non-spell healing you'll need to handle spell ID = 0 in the py script.

doug1234:
I saw there was an on heal received type event but that wouldn't really do what I wanted.  I can send out the "Healing Bonus" query from non spell healing with an id of zero.  Then I'll make sure I do nothing when the id is zero in the Augment Healing feat (I assume it would break otherwise).

BTW:  Is there a way to test for a valid spell id?  That would be the best way to handle things in Augment Healing.

DudeMcDude:
IIRC the accepted way is to test spell enum nonzero.

doug1234:
Thanks.  I'll get it finished up later this week.

doug1234:
Should be good to go.

doug1234:
Wolverine's Rage feat.

DudeMcDude:
No, I'm on vacation till sept 29. October probably.

doug1234:
Added Versatile Unarmed Strike and some cleanup.

DudeMcDude:
The issue with duplicates should be fixed now, so the check for wild shaped condition should be superfluous. Also merged :)

doug1234:
Thanks!  I'll check it later this week.

PR #308, opened 2018-08-13
Easy implementation for encounter packs

IPDTFE:
By intercepting the call for random_encounters and modifying the encounter only if it's repeatable it becomes easy for modders to implement additional encounter packs. 

All you need is a phyton file with a function whose name starts with EncounterPack which returns a list of encounters to be added to the pool. The encounter logic regarding the map, the party, flags and whatever else a modder may be interested in can all be done in this file, before returning the list of encounters. I've added a naming convention to avoid having conflicting EncounterPacks names.

The original function is preserved, and untouched, it's called by my own script before checking whether to modify the encounters. The original encounters are also preserved, there's an equal chance for any of the given encounters to appear, challenge rating aside.

The only doubt I have is whether the directory specified at the line 17 of expand_encounters is correct for the final publication, or it needs to be modified accordingly.

I've also attached a testpack with a simple encounter against up to two atropal scions, it can be expanded as a guideline for modders or removed.


tpdatasrc/co8fixes/scr/expand_encounters.py
```diff
@@ -0,0 +1,29 @@
...
+def encounter_coordinator( setup, encounter ):
+	var = encounter_exists(setup, encounter) 						#original function is called immediately
+	if(get_f('qs_is_repeatable_encounter') and game.global_flags[500] == 1):		#checks if the encounter is a random and NC is enabled
+		extra_encounters(setup, encounter)						#therefore avoids interrupting scripted and special encounters
+	return var						
+		
+def extra_encounters(setup, encounter):
+	expanded_list = []
+	
+	encounterpacks = os.listdir('.\data\scr')	#correct location for the final release (?)
```
shartte: this will re-run everytime it checks for encounters. is that intended? shouldn't this actually be only done once?
in addition, we need to use TIO functions here since they should also use files from .dat files, not only files that are unpacked on disk.
could the result of this function be cached?

IPDTFE:
I'm still not sure whether my files should be moved and if I need to make any changes regarding caching and TIO's functions. If anyone can point me in the right direction I'd appreciate it.

DudeMcDude:
Very nice, I'll review over the weekend.

IPDTFE:
I see, I will look into it. 

13,75~ approximated is the average number of encounters in the NC version of co8,  therefore if we pull a random number from 14 + how many encounters were added through encounter packs you have a balanced chance of having a random encounter from the co8 version or the added ones.

IPDTFE:
the files no longer rely on the co8 version, I've expanded the RE_class to accomodate a maplist.

 I've changed the files location, still not sure whether it's the correct one. Need to change the directory where files are found, but I'm not sure where we want that to be.

Tested both with co8 and GOG version, seems to work fine.

IPDTFE:
I'm open to suggestions, fixes and improvements. I have never worked with anything like this so far but I believe the program should work as it is. 

If there's something that needs to be worked on let me know, thank you.

DudeMcDude:
The file lookup should be improved, as you had noted. Also the magic number if possible ;)
Please remind me over the weekend.

IPDTFE:
I've made sure the new function is called only if we have a valid result while checking for an encounter.

Regarding expand_encounters, now encounter packs need to be added to a folder inside the main TOEE folder called encounterpacks, the name convention is still required in order to avoid conflicting names from different sources.

The magic number has been expanded to take in account sleep encounters and whether NC is enabled. It can still be improved by checking the exact number of sleep encounters depending on the map since it has a decent variance unlike wilderness encounters, I might add it in the future.

Below you'll find a test pack to be added to the folder encounterpacks in the main TOEE directory, let me know if anything else needs to be changed, added or moved.
[EncounterPackTest.zip](https://github.com/GrognardsFromHell/TemplePlus/files/2343342/EncounterPackTest.zip)

IPDTFE:
I'll look into changing how the number of base encounter is handled.

Regarding general path handling, do you mean the ability to choose the directory where files are accessed? I have no knowledge whatsoever of linux so I'm not sure where the issue is.

DudeMcDude:
/ is linux

IPDTFE:
I've been a little busy and I still am, I will need some more time before I can finish this.

PR #310, opened 2018-10-11
Added Powerful Charge, Greate Powerful Charge, Two Weapon Rend and Oaken Resilience Feats.  Changed improved unarmed strike to default to real damage instead of subdual.

doug1234:


DudeMcDude:
It seems you set the BAB requirement to +1 instead of +4?

doug1234:
Oops!  The +4 is wrong.  I'll change it to +1.

doug1234:
The +4 is for greater powerful charge which I will be adding shortly.

doug1234:
The default for monks was real damage.  Now unarmed strike works the same way (which makes sense to me).

It was only setting dice 0 before.  The way I did it would make sense for feats like powerful charge and such but not if you somehow were doing fire damage on you unarmed attacks.  I could set it back to only effecting dice zero or I could change it to only convert certain types of damage to subdual.  What do you think?

doug1234:
FYI:  I did some testing today and it looks like fire damage from weapons is accounted for differently and is not effected by my change.  I am leaning toward putting in a check for the damage type  (bludgeoning, piercing, or slashing) just in case there is some way to get a type of damage like fire or acid.  I can't think of a way to get it to happen though.

DudeMcDude:
IIRC there are two OnDamage phases, this may be going on the first.

doug1234:
So now only physical damage will be converted to subdual.  I also added a comment to clarify.  Hopefully the change is good to go now.

doug1234:
Ah yes...  I didn't check the disassembly.  Looks like the two weapon defense feat uses it too and I am going to have to make some changes.

doug1234:
OK.  It should work correctly in all conditions.

PR #318, opened 2018-11-12
Added Natural Bond feat.

doug1234:


doug1234:
Thanks for the clarification.  I think I understand what you are looking for now.  I could add a flag to feat.h (FPF_POWER_CRIT_ITEM maybe) but it would be very specific to power critical and not reusable by another similar feat like Weapon Supremacy since the prereqs would unfortunately be different.

DudeMcDude:
You could use a script for that, no?

doug1234:
Normally it is one script per child feat.  I could maybe work around that but it would still need to be hooked up.  I don't see how to make it completely generic.  I can make the change to use the flag and see how that works.

doug1234:
Now keying off a flag in feats.h.  Let me know what you think.

doug1234:
I've made some good progress on the elven subraces but want to put that in a new pull request.  Any reason this one can't be moved to the main branch?

PR #329, opened 2019-02-23
 No longer allowing sneak attack against concealed creatures, telling blow support for scout and freedom armor

doug1234:


doug1234:
Of course not!  I don't make copy past errors.  The favored scout class is my favorite class in 3.5 edition.

I'll get that fixed Monday. :)

DudeMcDude:
Hi, can you update the dat files?

doug1234:
I'll take care of that later this week.  I also have one more thing I would like to add to the pull request.

doug1234:
Should be ready to go now!

PR #332, opened 2019-04-20
Added warmage

doug1234:
Help coming soon.


doug1234:
Good suggestion.  I think I can make that work if you can add the same condition to a creature multiple times (with different spell ids as args).  Since a creature could get hit with two different ice storm spells and be taking damage from melf's acid arrow in the same round for example.  Is it possible to add the same condition multiple times to a creature?  Do you have to set anything special to allow the condition to exist multiple times on the same creature?

DudeMcDude:
Yes, you can add multiple times - there's an arg in PythonModifier to add a callback that prevents multiples, which is true by default. You just need to set it to false.

doug1234:
Thanks!  I'll give that a try.

doug1234:
Changed the code to work the way you suggested.  It was a bit more complicated though and I had to add a special case to get warmage damage for acid arrow on the second round.

DudeMcDude:
What's the issue with acid arrow?

doug1234:
On the second round, acid arrow damage is triggering before the War Mage Edge Damage condition from the first round is removed.  I want to make sure acid arrow gets the bonus each round.

PR #339, opened 2019-06-08
Beguiler

doug1234:
Added Beguiler.  Help files coming later.


DudeMcDude:
Very nice!

I see you've made a slight alteration to the chargen flow. Please make sure it doesn't cause any unwanted behavior when the player goes back a step in the chargen stages. (not saying it does so, just being careful)

doug1234:
Thanks!  I checked going back a step but I'll give it another look when I get a chance to make sure it can't get messed up.

PR #345, opened 2019-07-05
Added Battle Caster.

doug1234:


doug1234:
I am not sure what you mean by generic queries.  Do you mean that the query should have a more generic name?  I couldn't come up with anything better than the name I gave it.  Any suggestions?

DudeMcDude:
I mean that instead of checking an explicit list of classes with armored casting, the classes should have a property designating them as such.

doug1234:
Interesting idea but how would I make it work correctly when taking the first level of a class?  For example if a level 2 fighter takes a level of bard, they should be able to take the feat that level.  I don't think that would work correctly using a query.

DudeMcDude:
Yeah I didn't mean d20query, just a bool property you can query via the D20classSys.

doug1234:
OK.  I get it now.  I'll make an update some time next week.

doug1234:
Should be good to go now.

doug1234:
The list is now compiled only one time now.

PR #346, opened 2019-07-07
poison improvements

anatoliy-savchak:
* reimplemented PoisonedOnAdd
* implemented Unconscious to effects
* implemented paralyse, unconscious in delayed effects;
* fixed bonus in delayed effect;
* added poisons.json load poison specs


shartte:
Thanks a lot for your contribution!

@DudeMcDude probably needs to properly review, but I did a quick read over and it generally looks good to me. I have a few minor nitpicks which are not blockers:
- Error messages in the JSON loading code still reference "text styles" (probably copy pasta)
- I think we could get rid of the `poisonSystem` global variable and replace with `gameSystems->GetPoison()` where it's used.
- The "GetPoisonSpec" method currently checks whether `poisonSystem` is null, and if so, loads from the vanilla poison specs, but this likely is unnecessary (can it ever be called before PoisonSystem is initialized?)

anatoliy-savchak:
Thanks @shartte .
Say I fix these issues, then what? I'm kind of new to github. Should I commit and pull another request, or this pull request will contain new commit (as in bitbucket)?

shartte:
@anatoliy-savchak It works the same as in Bitbucket, if you push any changes to the branch in your repository, it'll automatically update this pull request. Also, don't feel pressured to fix any of these nitpicks, it's minor stuff that can also be fixed after merging.
I'll also try to fix the "all checks have failed" stuff because that's actually on me, heh.

anatoliy-savchak:
Sitra asked me to include poisons.json into tpdatasrc so I still going to do a new commit.

Most likely I will include vanilla poisons too. To have lookups in ToEEZ.ProtoEd editor.

anatoliy-savchak:
Committed new changes:
* Added PoisonEffect enum;
* Added poisons export SaveSpecsFile
* replaced global var poisonSystem with gameSystems->GetPoison();
* Fixed parse messages;
* Fixed dice.bonus parser to use int not unsigned int to allow minuses;
* Added default poison effect to be -10 (None) in Poison Spec loader;
* Added TemplePlus\tpdatasrc\tpgamefiles\mes\poisons.json with vanilla specs included;

DudeMcDude:
Looks good
A few things:
1. poison.json should be in /rules not /mes
2. For it to take effect you need to zip it into tpgamefiles.dat
( this will probably be redone anyway by doug i guess)
3. Have you verified that these two funcs are the only refs to poison specs?

anatoliy-savchak:
1. Will fix;
2. I'm not eager to touch dat files this time, can one else do it?
3. Yes, I searched all references before that in IDA, only in these two places.

anatoliy-savchak:
added new commit:
* moved poison.json from mes to rules;

PR #349, opened 2019-07-14
Natural attack improvements

anatoliy-savchak:
* Made attack 4 account now;
* Attack 4 now accounts size modifier;
* Weapon Finesse workaround to skip str negative bonus;

### DIFF
--- a/TemplePlus/condition.cpp
+++ b/TemplePlus/condition.cpp
@@ -476,23 +476,6 @@ public:
 		redirectCall(0x100FC18C, AnimalCompanionLevelHook);
 		redirectCall(0x100FC2D6, AnimalCompanionLevelHook);
 		redirectCall(0x100FC6D4, AnimalCompanionLevelHook);
-
-		// fix for negative str mod still accounting in the bon list
-		static int(*orgWeaponFinesseToHitBonus)(DispatcherCallbackArgs) =
-			replaceFunction<int(__cdecl)(DispatcherCallbackArgs)>(0x100F80C0, [](DispatcherCallbackArgs args)->int
-		{
-			auto dispIo = static_cast<DispIoAttackBonus*>(args.dispIO);
-			int bonCount = dispIo->bonlist.bonCount;
-			int result = orgWeaponFinesseToHitBonus(args);
-			if (bonCount != dispIo->bonlist.bonCount) // dex bonus added
-			{
-				int strMod = objects.GetModFromStatLevel(objects.abilityScoreLevelGet(args.objHndCaller, stat_strength, 0));
-				if (strMod < 0) {
-					dispIo->bonlist.ModifyBonus(-strMod, 2, 103);
-				}
-			}
-			return result;
-		});
 	}
 } condFuncReplacement;
 
--- a/TemplePlus/condition.cpp
+++ b/TemplePlus/condition.cpp
@@ -312,6 +312,54 @@ public:
 	void apply() override {
 		logger->info("Replacing Condition-related Functions");
 
+		//dispTypeConditionAddPre
+		static int(__cdecl* orgTempAbilityLoss)(DispatcherCallbackArgs) = replaceFunction<int(__cdecl)(DispatcherCallbackArgs)>(0x100EA1F0, [](DispatcherCallbackArgs args) {
+			Stat statDamaged = (Stat)args.GetCondArg(0);
+			int amountDamaged = args.GetCondArg(1);
+			DispIoCondStruct *dispIo = dispatch.DispIoCheckIoType1((DispIoCondStruct *)args.dispIO);
+			if ((int32_t)dispIo->condStruct == args.subDispNode->subDispDef->data1 && dispIo->arg1 == statDamaged){
+				int amountDamagedByNew = dispIo->arg2;
+				int scoreLevel = objects.StatLevelGet(args.objHndCaller, statDamaged);
+				if (scoreLevel - amountDamagedByNew < 0){
+					amountDamagedByNew = scoreLevel;
+					if (amountDamagedByNew <= 0)
+					{
+						dispIo->outputFlag = 0;
+						return 0;
+					}
+				} 
+				amountDamaged = amountDamaged + amountDamagedByNew;
+				args.SetCondArg(1, amountDamaged);
+				dispIo->outputFlag = 0;
+				critterSys.CritterHpChanged(args.objHndCaller, objHndl::null, 0);
+			}
+
+			return 0;
+		});
+
+		//dispTypeConditionAdd
+		static int(__cdecl* orgTempAbilityLoss2)(DispatcherCallbackArgs) = replaceFunction<int(__cdecl)(DispatcherCallbackArgs)>(0x100EA3B0, [](DispatcherCallbackArgs args) 
+		{
+			Stat statDamaged = (Stat)args.GetCondArg(0);
+			if ((statDamaged >= stat_strength) && (statDamaged <= stat_charisma)) {
+				int scoreLevel = objects.StatLevelGet(args.objHndCaller, statDamaged);
+				if (scoreLevel < 0){
+					int amountDamaged = args.GetCondArg(1);
+					// dont let stat go below zero
+					amountDamaged = amountDamaged + scoreLevel;
+					if (amountDamaged <= 0)
+					{
+						// remove condition, as it should have no effect
+						conds.ConditionRemove(args.objHndCaller, args.subDispNode->condNode);
+					} else {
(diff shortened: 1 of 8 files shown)


TemplePlus/dispatcher.cpp
```diff
@@ -19,6 +19,8 @@ class DispatcherReplacements : public TempleFix {
 public:
 
 	static int Dispatch54AoE(objHndl, DispIO *, D20DispatcherKey);
+	static int GetNumNaturalAttacks_(objHndl handle);
```
DudeMcDude: Why is this in dispatcher.cpp?? It has nothing to do with dispatchers...

Usually you can use the raw address to find a better place for it.

In this case it's obviously part of critterSys.

TemplePlus/dispatcher.cpp
```diff
@@ -72,6 +77,40 @@ int DispatcherReplacements::Dispatch54AoE(objHndl handle, DispIO* evtObj, D20Dis
 	return 0;
 }
 
+int DispatcherReplacements::GetNumNaturalAttacks_(objHndl handle) {
+	if (!handle) return 0;
+	auto obj = objSystem->GetObject(handle);
```
DudeMcDude: For temple+ please also check obj.
Although rare, we sometimes get invalid non-null handles, which return a null object.

TemplePlus/dispatcher.cpp
```diff
@@ -72,6 +77,40 @@ int DispatcherReplacements::Dispatch54AoE(objHndl handle, DispIO* evtObj, D20Dis
...
+	int result = 0;
+	int idx = 0;
+	int attacksCount;
+	do
+	{
```
DudeMcDude: do{} while is bad form. It's an artifact of the decompiler, there's no need to adhere to it.

TemplePlus/dispatcher.cpp
```diff
@@ -72,6 +77,40 @@ int DispatcherReplacements::Dispatch54AoE(objHndl handle, DispIO* evtObj, D20Dis
...
+	return result;
+}
+
+int DispatcherReplacements::GetDamageIdx(objHndl handle, int idx)
+{
+	if (!handle) return 0;
+	auto obj = objSystem->GetObject(handle);
+	int v2 = 0;
+	int result = 0;
+	while (1)
```
DudeMcDude: Likewise, decompiler produces crappy formatting. What seems like an infinite loop is really just a simple for loop.

int __cdecl GetDamageIdx(ObjHndl ObjHnd, int attackIdx){
int attackIdxSum = 0;
for (auto idx = 0; idx < 5; idx++){
    attackIdxSum += obj->GetInt32(obj_f_critter_attacks_idx, idx);
    if ( attackIdxSum > attackIdx )
      return idx;
  }
  return 0; // attackIdx is bigger than the sum of all obj_f_critter_attacks_idx items

TemplePlus/critter.cpp
```diff
@@ -1810,6 +1815,56 @@ void LegacyCritterSystem::BuildRadialMenu(objHndl handle){
 		}
 	}
 }
+int LegacyCritterSystem::SetCritterAttacks(objHndl handle)
```
DudeMcDude: This belongs in protos.cpp

TemplePlus/dispatcher.cpp
anatoliy-savchak: I wanted to change as little as possible. Thanks for advice. Will fix

TemplePlus/condition.cpp
```diff
@@ -476,6 +476,23 @@ class ConditionFunctionReplacement : public TempleFix {
 		redirectCall(0x100FC18C, AnimalCompanionLevelHook);
 		redirectCall(0x100FC2D6, AnimalCompanionLevelHook);
 		redirectCall(0x100FC6D4, AnimalCompanionLevelHook);
+
+		// fix for negative str mod still accounting in the bon list
+		static int(*orgWeaponFinesseToHitBonus)(DispatcherCallbackArgs) =
```
DudeMcDude: One of my life's biggest regrets is this long-ass file :)
Please put it in ability_fixes.cpp

TemplePlus/critter.cpp
anatoliy-savchak: Funny, I first created in protos, but then tried to match the pattern, therefore moved here :)

Will fix, thanks

TemplePlus/dispatcher.cpp
DudeMcDude: Also, it looks like I already rewrote this function!

And in critterSys too :)

DudeMcDude: On that note. Since Temple+ already replaces a whole bunch of stuff, you should be extra careful when replacing existing functions. Sometimes doing a replaceFunction won't help because of exactly what happened here: I rewrote this function and called it explicitly in the code. That's my fault really, but that's how it is right now...

anatoliy-savchak:
Added new commit with such changes:
* Moved SetCritterAttacks from critter.cpp to protos.cpp;
* Moved GetNumNaturalAttacks from dispatcher.cpp to critter.cpp;
* Updated GetNumNaturalAttacks to use "for" loop;
* Deprecated GetDamageIdx hook as already implemented;
* Moved orgWeaponFinesseToHitBonus from condition.cpp to ability_fixes.cpp;

anatoliy-savchak:
No, its fine: from 0 to 3, therefore <4. See original:
  do
  {
    v3 = getArrayFieldInt32(obj, obj_f_critter_attacks_idx, v2);
    if ( v3 > 0 )
      v1 += v3;
    ++v2;
  }
  while ( v2 < 3 );

anatoliy-savchak:
image

TemplePlus/protos.cpp
```diff
@@ -431,3 +435,54 @@ int ProtosHooks::ParseSpell(int colIdx, objHndl handle, char* content, obj_f fie
...
+	int32_t objType = obj->GetInt32(obj_f_type);
+	int result = 0;
+	if (objType == 13 || (result = objType, result == 14))
+	{
+		int32_t strAbility = obj->GetInt32(obj_f_critter_abilities_idx, 0); //v2 = getArrayFieldInt32(obj, obj_f_critter_abilities_idx, 0);
+		int strMod = objects.GetModFromStatLevel(strAbility);
+		int32_t sizeCategory = obj->GetInt32(obj_f_size);
+		int sizeMod = critterSys.GetBonusFromSizeCategory(sizeCategory);
+		int attackIndex = 0;
+		do
```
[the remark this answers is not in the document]
DudeMcDude: Please fix this

Also please us obj->IsCritter() instead of explicit type==13 || 14

TemplePlus/protos.cpp
```diff
@@ -431,3 +435,54 @@ int ProtosHooks::ParseSpell(int colIdx, objHndl handle, char* content, obj_f fie
...
+				auto dicePacked = dice2.ToPacked();
+				obj->SetInt32(obj_f_attack_bonus_idx, attackIndex, attackBonusNew);
+				obj->SetInt32(obj_f_critter_damage_idx, attackIndex, dicePacked);
+			}
+			++attackIndex;
+		} while (attackIndex < 3);
+
+		auto dexAbility = obj->GetInt32(obj_f_critter_abilities_idx, 1);
+		result = objects.GetModFromStatLevel(dexAbility);
+		for (int i = result; attackIndex < 4; ++attackIndex)
```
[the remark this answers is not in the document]
DudeMcDude: This too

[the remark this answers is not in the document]
anatoliy-savchak: Set Critter Attacks is handy function to remove str from damage dice and from attack bonus, because it will later be added during actual attack.

But the last 4th natural attach seems to be reserved specifically for Weapon Finesse attacks. It only removes dex mod from attack bonus. But it does not remove str mod from damage dice. Which is odd, as it still is going to affect damage dice.

**I suggest add str damage bonus removal similarly to 3 first natural attacks.** 

Arguments:
1) 4th attack is almost never used in protos. Exception - one dragon and one dude.
2) 4th attack even if it was specified in the protos was actually never used in vanilla (thus one of fixes above);
3)  Let's say I have Small Centipede with 1d4-3 damage. Without this fix I will need to remember that in 4 attack I need to manually change -3 to 0 (to simply 1d4), otherwise in reality during damage it will be 1d4-6 (as str mod is -3).

DudeMcDude:
Maybe it's meant for rangef attack?

anatoliy-savchak:
> Maybe it's meant for rangef attack?

I have not thought about that. OK, then I will not change it. Thanks!

PR #351, opened 2019-07-24
 Fixed memorized spell slots showing up for new natural casting classes when the character has more than one casting class.

doug1234:


doug1234:
The original function is being called which already takes care of hiding the widgets for bards and sorcerers.  It only needs to be done for the new classes.

PR #353, opened 2019-07-31
 Fixed an issue where vancian cast spells were not getting removed from the available list if a sudden feat was used (due to the metamagic data being different).

doug1234:


doug1234:
It was indeed a bad init.  I removed the workaround fix and just added a better one.

PR #364, opened 2019-09-01
Fix D20 Attack Power enumeration

shartte:
For example uses see Vanilla @ 0x100faac0 (Ki Strike) or 0x100e0e00, where damage type is converted into attack power (the more relevant location).

Also see 0x100e15c0, where the conversion for protos.tab is happening. It also translates flags like "Normal" to 0, "Unspecified" to 1, and so on.


DudeMcDude:
I think we should also tackle the Python constants (since they are the source of the error)

It seems like all the scripts use D20DAP_UNSPECIFIED  or D20DAP_NORMAL (so in practice, they're using UNSPECIFIED and SILVER), so it should be safe.

PR #376, opened 2020-03-03
Fix human and half-elf favorite class

edoipi:
Human and half-elf favorite class were not accouting for new
base classes, for example Barabarian 1/Fighter 1/ Scout 3 was
getting penalties.


doug1234:
Had time to checkout the branch and test this today and it works great.  Since there have been no other comments, I am going to go ahead and merge it.

DudeMcDude:
Oh oops, I didn't even notice this was a PR (I thought it was a bug report).

I see you've added human and halfelf modifiers, do they include everything they should?

edoipi:
I certainly hope so. If I remember correctly I didn't include human additional skill in its definition, but testing showed that skill points were still applied, so I assumed it is hardcoded somewhere deeper in the game.

DudeMcDude:
Unfortunately this breaks half elves, please fix.

edoipi:
Well that sucks. Can you specify what you found wrong? Back then when I made this change I tested half-elf(on a new character only) and everything seemed fine.

DudeMcDude:
Missing import char_class_utils in half_elf

doug1234:
I went ahead and made the fix in my pull request.  Do you think this is worth releasing a new version early?  I had planned to release a new version at the end of the month but could push one out Wednesday with what we have currently.

DudeMcDude:
Yes, please do so.

PR #380, opened 2020-03-26
game.make_custom_name and new tactics

anatoliy-savchak:
* added game.make_custom_name
* new tactic goto added;
* new tactic halt added;
* new tactic stop added;
* new tactic target obj added;
* new config disableChooseRandomSpell_RegardInvulnerableStatus added;
* re-added get_handle_upper and get_handle_lower;


anatoliy-savchak:
New function game.make_custom_name(new_name) is quite useful to assign names to dynamic objects from Python, without altering mes files.

New config.disableChooseRandomSpell_RegardInvulnerableStatus is essential for controlling spellcast able NPC, as current AI implementation will ignore any strategy defined and will go AWOL. I'm pretty sure it can be redesigned in better way, not sure how though.

goto and target obj rely on middle string in tactics triplet, which was unused before. I tested it, works just fine.

See this file: https://github.com/anatoliy-savchak/ToEE.Cormyr/blob/dev/data/scr/utils_tactics.py
Examples:
`
tac = utils_tactics.TacticsHelper("banelar")

tac.add_target_obj(game.party[4].get_handle_lower(), game.party[4].get_handle_upper())
tac.add_attack()

tac.add_target_self()
tac.add_cast_single(utils_spell.spell_name(spell_invisibility), class_wizard, caster_level)
tac.add_goto(497, 486)
tac.add_use_potion()
tac.add_stop()
`

The halt and stop are designed to do do following:
halt would remove move action;
stop would end any action left;

anatoliy-savchak:
@DudeMcDude I saw your comment about handle upper/lowe in pull request 371 from @doug1234, and I agree that it somewhat bad form to expose actual handles.

But **target obj** would greatly change and expand AI modding, and would grant great enjoyment to the game. And we have only two int32 fields to use. So perhaps you could favour the API after all.

shartte:
@anatoliy-savchak Exposing handles in such a way will make it impossible to actually change how handles are represented internally. Also: handles do not survive save/load and will lead to inexplicable crashes if used incorrectly. That's the primary reason one should prefer to use object IDs rather than handles. If there aren't enough APIs in Python to work with object IDs, THAT should be fixed instead.

anatoliy-savchak:
@shartte, sure - I'm good with that objection.
How about such way of obj.id?

```
{"id", PyObjHandle_GetID, NULL, NULL },

static PyObject* PyObjHandle_GetID(PyObject* obj, void*) {
    auto self = (PyObjHandle*)obj;

    if (!self->handle) {
        return PyString_FromString("OBJ_HANDLE_NULL");
    }
    else {
        auto name = self->id.ToString();
        return PyString_FromString(name.c_str());
    }
}
```

And lookup for the actual handle during middleString parsing. 
Would it be ok then?

shartte:
@anatoliy-savchak I had to go back and check the code. I don't think you actually _need_ to add anything, it should just work. You can simply hold onto a game object within Python. There is no need to convert it to a string. Do you want to save these objects to the save game or survive map changes?

anatoliy-savchak:
@shartte 
> There is no need to convert it to a string. Do you want to save these objects to the save game or survive map changes?

Unfortunately - no. I really need to convert it to string.

There are two usage of obj.id:
- Pass it to middle triple of obj.ai_strategy_set_custom(["aaa", "target obj", "**the_id**, ""]);
- Uniquely create file name under save\savename\objects\ **the_id** .json;

The most trivial way to implement first was to use the handle. It's wrong to expose, yes I agree.
Second way would be using either str(PyHandle) or new property - id. But str() would have npc description, which actually can change, so...

shartte:
You will suuuuuuuuuuuper crash the game if you do that with handles :-D
If you change maps, object handles become invalid.

What you can do without extra code:
You can pickle and unpickle game objects using the standard Python pickle tools: https://docs.python.org/2/library/pickle.html 
You can do this without actually having to know how ToEE serializes/deserializes the object handle.
You can also pickle to/from strings.

That is what ToEE will do internally for arguments to python timers.

Can you try this?
```
import pickle # (at the top)

some_obj_str = pickle.dumps(some_obj)
.......
# at a later date
.......
some_obj = pickle.loads(some_obj_str)
```

anatoliy-savchak:
@shartte, thanks, I will investigate pickle :)

And I agree that using handles is not good. Fine.
Can you take a look at commit I just made? (tactic target obj via obj.id 825f5141e544860670046bc4eb44de51130d31f3) for this pull request?
Would it be OK then?

anatoliy-savchak:
@shartte, as for the pickle implementation in PyObjHandle_getstate() => It does not work well for doors for example. Why?

Due to it is Positional, not Permanent. I guess changing from IsPermanent() to IsPersistable() in the PyObjHandle_getstate() would work. But it needs testing etc.

And it is not the subject of my pull request. **I just need to pass Object to tactics in safe way**.

shartte:
Oh wow, I just saw it. No one will understand that "re-use" of existing fields for other purposes in just a few weeks. But I am not going to tell you what to do ;-)
@doug1234 What do you think about this?, Especially the changes in ai.cpp

anatoliy-savchak:
Look guys, I need tactics expanded, and at least some sort of argument 0 and 1, as condition has. 

The pad2 and pad3 fields even if used, then only for specific tactics, where spell is present. Otherwise it will always be just allocated memory.

I also want to add more tactics, for example ideas:
- touch attack;
- single attack with weapon;
- secondary attack with bite (as in MM);
- total defense, wouldn't it be great? :)
- sneak (!!)
- throw (grease or smoke stick, simply great)
- sunder (I would love to give pc some pain))
- open lock (target - door with animal for example)
- use potion by obj (not randomly as it is now)
- use potion by name_id (not randomly as it is now)
- open container (door maybe, need to check)

@shartte, you did great job pointing me on incorrect usage of handles - thank you! I mean it :)

If question is only by pad2\pad3, it could be addressed later by re-implementing "correct" way. Implementation still is going to be encapsulated, so... But API usage would be the same.

API
`attachee.ai_strategy_set_custom.['banelar-attack', 'target obj', 'G_BAE066DE_A0A2_4078_8AF0_F0A26A5434FA', '', 'attack', '', '']`

doug1234:
The code definitely is not obvious.  I suggest adding some comments to state what is going on in ai.cpp.

anatoliy-savchak:
@shartte, @doug1234, @DudeMcDude  - added new commit with comments and faster looking up objects by id.

Importance of this new tactics I'm explaining in this video, which I created specifically to persuade you to approve this change. Thank you!

https://www.youtube.com/watch?v=MGLt6HVT0zw

doug1234:
Thanks for making the change.  I plan to merge it in shortly unless anyone comments.

PR #382, opened 2020-04-08
game.combat_turn and Path constructors fix

anatoliy-savchak:
* pathfinding.h - better fix of PathQueryResult.nodeCount not initializing;
* python_game.cpp - added game.combat_round Python API;
* ui_intgame_turnbased.cpp - remove obsolate debug print;
* animgoals_callbacks - fixed IsConcealed on non-critter;
* traps.h - figured out  unknown fields in Trap;

### DIFF
--- a/TemplePlus/animgoals/animgoals_callbacks.cpp
+++ b/TemplePlus/animgoals/animgoals_callbacks.cpp
@@ -84,7 +84,7 @@ int GoalIsProne(AnimSlot& slot)
 int GoalIsConcealed(AnimSlot& slot)
 {
 	//logger->debug("GoalState IsConcealed");
-	return (critterSys.IsConcealed(slot.param1.obj));
+	return (objects.IsCritter(slot.param1.obj) && critterSys.IsConcealed(slot.param1.obj));
 }
 
 // Originally @ 0x100125F0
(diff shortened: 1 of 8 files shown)


doug1234:
Looks pretty uncontroversial to me.  I am going to go ahead and merge it in.

DudeMcDude:
Why did you remove the debug print? What's obsolete about it?

anatoliy-savchak:
> Why did you remove the debug print? What's obsolete about it?

I assume debug print for cursor changing was very handy during cursor fixes. But now when it is polished, it spams debug output to the point of complete unusability. Sometime I cannot see nothing but hundreds of "cursor changed" notifications.

I was hesitant to make this change, as someone added it to help test cursor changes. That is nice, and I honor it. But now it works stable and there is no necessity to leave it, that is why it is obsolete.

DudeMcDude:
It's strange that you see it toggle so much. I don't recall seeing anything like that. It could be an actual problem you're glossing over. Why would it change so many times?

anatoliy-savchak:
I guess it is related to moving mouse from game to devenv on other monitor to see logs, then back again and so forth. It flies over cursor-changing objects and spams debug window.

I can uncomment it if you like. No issue here.

DudeMcDude:
In general, we need all the help we can get when figuring out user submitted bug reports. I guess this one isn't so important, but you never know. I'd rather keep it.

PR #384, opened 2020-04-11
Several minor changes before release of the next version

doug1234:
Several minor things found during testing and some graphics tweaks.  This is hopefully my final pull request before the next release.


TemplePlus/temple_enums.cpp
```diff
@@ -455,8 +455,13 @@ static const char *ObjTypeNames[] = {
 	"obj_t_bag" // 16
 };
 
+static const char* unknownObject = "obj_t_unknown";
+
 const char* GetObjectTypeName(ObjectType type) {
-	return ObjTypeNames[type];
+	if (type < std::size(ObjTypeNames)) {
```
shartte, line 461: Doesn't this just hide other bugs?

TemplePlus/ui/ui.cpp
```diff
@@ -790,6 +790,9 @@ bool UiManager::TranslateMouseMessage(const TigMouseMsg& mouseMsg)
 			bool enqueueExited = false;
 			auto globalWid = GetWidget(globalWidId);
 			// if window
+
+			if (globalWid == nullptr) return false;  //Happens sometimes on shutdown
```
shartte, line 794: This probably also hides other bugs :-D

doug1234: This only happened on shutdown when TOEE is in windowed mode.  At this point it is probably OK that the window is NULL (it needs to get destroyed).  The mouse handler just needs to return.  This change prevents the crash and TOEE seems to shutdown successfully.

TemplePlus/temple_enums.cpp
doug1234: Thanks for taking a look at the code.  This hopefully does the opposite.  I was seeing a crash in a debug message (an entity with no inventory was getting its inventory checked I think) and this change allows the message to be printed instead of crashing (the issue really wasn't about object types it was just trying to print out the object type for an entity with no inventory in a debug message resulting in a crash instead of a message).

TemplePlus/ui/ui.cpp
DudeMcDude: If it's destroyed normally then it's better to directly check that the windows have been destroyed.

TemplePlus/temple_enums.cpp
DudeMcDude: Why did this start happening all of a sudden?
You have to tackle issues like this at the root.

TemplePlus/ui/ui.cpp
doug1234: Sounds good.  How do you check for a destroyed window (other than check for null pointer)?

TemplePlus/temple_enums.cpp
doug1234: This happen when I mess around testing spells in the starting room.  I assumed I was just triggering one of the many debug messages that you get while running (just this debug message was crashing TOEE).  I am not against eventually chasing down whatever is causing this but I doubt it is critical.  The unprotected message was however making things worse.  Now it should log the issue.

TemplePlus/ui/ui.cpp
doug1234: Just looked at the code.  It is checked for -1 before it gets into that part of the code.  It looks like GetWidget() just didn't find in in its widget list.

shartte: If this is a spurious mouse move event triggering this after the UI has already been destroyed, the actual bug might be not unregistering the mouse move handler on the mainwindow.
I specifically mean MainWindow::SetMouseMoveHandler and the other event handling functions. You're right that these callbacks may trigger after the UI has already been destroyed, so they likely need to be unregistered.

TemplePlus/temple_enums.cpp
shartte: The danger here is that in reality you're getting a corrupted object from somewhere, and it's obj_f_type is garbage because of the corruption. By just catching the conversion here, you'll never actually find the root cause for the corruption.

TemplePlus/ui/ui.cpp
doug1234: I think you are right and the cause probably is a spurious mouse move event.  However the handler gets created in the game view constructor and destroyed in the destructor so that looks ok.  Maybe game view is destroyed too late (it is destroyed at the end of GameLoop::Run()).

TemplePlus/temple_enums.cpp
doug1234: I can look into this one a bit more and see if I can fix the underlying cause (or atleast add a bug report to fix later).  I should probably have some time for that on Tuesday.  I do still think this change is for the better though.  Log messages that can cause undefined behavior are very bad IMHO.  If the code went whack-a-do later, we would still have the log message to figure it out.

TemplePlus/ui/ui.cpp
doug1234: Here is what the stack looks like.  It seems that the game loop has actually already been destroyed and the handler has been set to null.  Maybe The UiPcCreationSys destructor should check for a null game loop and not call mRace->Reset(uiPcCreation.GetCharEditorSelPacket()); if that is the case?

image

doug1234: My suggestion above seems like it would indeed fix the problem but I am still not 100% sure it is the most correct solution.

TemplePlus/temple_enums.cpp
doug1234: Looks like ObjRegistry::Get() is returning a bad handle.  If I change the method back to using the "old way", it returns a null pointer instead of a bad pointer.  The crash happens when loading a saved game after the game has been started and one of the new spells is cast.  It seems that the particle system for the new spell effect is causing the query.  Do you know how the removal of the particle system happens for spells?

doug1234: Do you know how TOEE removes all the spell effects on shutdown?  Seems that an effect is getting removed too late in the shutdown process.  I am having some trouble tracking down how they actually get removed.

TemplePlus/ui/ui.cpp
shartte: What we're actually missing here is something else, heh:
When the widget that `mWidgetMouseHandlerWidgetId` has the id of is removed, it should clear the field.
I.e.:

In `void UiManager::RemoveWidget(LgcyWidgetId id)`, the code should do something along those lines (at the function's end):
```
UnsetMouseCaptureWidgetId(mWidgetMouseHandlerWidgetId);
if (mMouseButtonId == id) {
mMouseButtonId = 0;
}
if (mWidgetMouseHandlerWidgetId == id) {
mWidgetMouseHandlerWidgetId = 0;
}
```

doug1234: This change may well be for the better but it doesn't fix the crash I am seeing.  When make your change and remove my fix, TOEE can still easily be crashed by pressing the "x" at the start menu.

shartte: I'll look into it if it that's easy to repro :-D

TemplePlus/temple_enums.cpp
doug1234: I now have the root issue fixed in my latest push.

TemplePlus/ui/ui.cpp
doug1234: If you find a better way to prevent the crash on shutdown I think we can add it in a new pull request.  I don't think there is much more to say on this pull request so I am going to go ahead and merge it in.

DudeMcDude:
Please revert the changes in ui and temple enums now that you've resolved the root causes.

doug1234:
Looks like there was a bit of confusion.  Here is a summary of where things stand:
(1)  The UI stuff has not changed.  shartte is looking into a better UI fix but the way it is now is better than before.  We can revert when he has a final fix but I don't see it as necessary for this version.
(2)  I fixed the root issue of the other problem.  However, I would still prefer a log message to a crash.  If we really want it to stop there I guess I could change it to a log message and an exit, but I don't know if getting a bad object always represents a critical error.

PR #385, opened 2020-04-23
tpdp.dispatch_skill and other Python API changes

anatoliy-savchak:
* common.h - added constructors for BonusEntry and BonusCap;
* poison.cpp - added history line when damage is dealt for ability loss. Was unclear;
* python_dispatcher.cpp - added dispatch_skill to get bonuse list;
* python_dispatcher.cpp - added create_history_type6_opposed_check, tested thoroughly;
* python_dispatcher.cpp - added EventObjD20Signal.get_d20_action, see ActionSequenceSystem::ActionBroadcastAndSignalMoved()
* python_game.cpp - added game.alert_show;
* python_integration_encounter.cpp - added python error print for can_sleep;
* ui_alert - UiAlert::ShowEx implementation;

### DIFF
--- a/TemplePlus/common.h
+++ b/TemplePlus/common.h
@@ -100,6 +100,13 @@ struct BonusEntry
 	int32_t bonType; // types 0, 8 and 21 can stack ( 100E6490 ); use negative number for modifier
 	const char * bonusMesString; // parsable string for the help system e.g. "~Item~[TAG_ITEM]"
 	char * bonusDescr; // e.g. "Magic Full Plate +1"
+
+	BonusEntry() {
+		this->bonValue = 0;
+		this->bonType = 0;
+		this->bonusMesString = nullptr;
+		this->bonusDescr = nullptr;
+	}
 };
 
 struct BonusCap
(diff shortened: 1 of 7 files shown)


doug1234:
Looks good to me.  I may wait on getting this version out first unless you really need these changes.  BTW:  Why was the Show() method non-static but ShowEx() was made static?

anatoliy-savchak:
Hi Dough

I assume Show method was not finished and left as non-static. And I did not wonted to touch code that I have not written.

As for the release...
I have created "Napping" condition, which allows game designers to set a Critter into "Sleeping" mode. Btw it looks pretty cool when I was testing that.

The code - https://github.com/anatoliy-savchak/ToEE.Cormyr/blob/dev/overrides/scr/tpModifiers/napping.py

The code relies on:
* tpdp.dispatch_skill()
* tpdp.create_history_type6_opposed_check()
* tpdp.EventObjD20Signal.get_d20_action()

Meaning that it wont work. At least with latest implementation.

The module I'm working on - Shattered Gates of Slaughtergarde.pdf has many levels to come. And I think it will take months to develop it. I'm currently only half-way through Slaughtergarde Laboratory map.

Therefore - yes, I agree, upcoming release could go without this commit. Although code change is not risky. So it won't affect main features and fixes.

doug1234:
Sounds cool.  I think I will have use for some of the functions you wrote also.  I will merge this in when we get the current release out unless anyone else has an issue.

doug1234:
I am going to go ahead and merge this one in.  If we need a bug fix release, this can probably be released with it without much concern.

Issue #233, opened 2016-04-09
[Bug] Random map encounter crash after having summoned & rested


TheMadSword:
Hi,

It seems possible to still crash due to map encounter.

1) Load the save
2) Level up the sorcerer : Pick Summon Monster 2, 3, 4,5
3) Get out of the building
4) Summon Monster 5 : Shadow Mastiff, make it move
5) Summon Monster 5 : Beareded Devil make it move
6) Go back in the inn
7) Rest 8h
8) Go out and map travel far away and hope to get an encounter, you should crash

Crash occurence : 3 out of 5 test

http://www.filedropper.com/save831randomencountercrashv2

PR #387, opened 2020-04-25
Worldmap fixes

DudeMcDude:
@doug1234 
That's one of the WIP fixes I'd wanted for the next release.
IIRC there were two issues actually:
1. Travelling to one of the new Co8 locations from a random encounter map, when it's not the original destination of travel. See thread:
https://co8.org/community/index.php?threads/release-dllfix11.3940/
I think that's fully fixed.
2. Crash due to getting data from the terrain bitmap wrong. See thread:
https://co8.org/community/index.php?threads/terrain-bmp.11642/#post-148308
I think I fixed the immediate crash issue by reploacing the GetTerrainType func and making it safe. But I think I was having some trouble calculating the offsets into the bmp data in memory. As it is, Co8 ignores it anyway and just randomizes the terrain, but it'd be nice to make the worldmap bitmap great again.

### DIFF
--- a/TemplePlus/fade.h
+++ b/TemplePlus/fade.h
@@ -6,6 +6,7 @@ enum FadeAndTeleportFlags : uint32_t {
 	ftf_advance_time = 8,
 	ftf_play_sound = 0x10,
 	ftf_unk20 = 0x20,
+	ftf_unk200 = 0x200,
 	ftf_unk80000000 = 0x80000000
 };
 
--- a/TemplePlus/gamesystems/legacysystems.cpp
+++ b/TemplePlus/gamesystems/legacysystems.cpp
@@ -2064,6 +2064,27 @@ const std::string &RandomEncounterSystem::GetName() const {
 	return name;
 }
 
+int RandomEncounterSystem::GetTerrainType(int x, int y)
+{
+	auto result = 0;
+	auto bitmapW = temple::GetRef<int>(0x109DD338 + 0x50C);
+	auto bitmapH = temple::GetRef<int>(0x109DD338 + 0x510);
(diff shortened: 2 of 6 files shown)


doug1234:
I am going to take another look at the first fix but it seems reasonable to me.

For the bitmap stuff, what are the magic numbers 528 and ?  The image is 264 x 283.  Also, what is up with shifting the image?

DudeMcDude:
Right shift is basically divide by 2.

doug1234:
I understand that you can integer multiply and divide by powers of 2 using left and right shift.  But I mixed up my left and right!  My question was more why does that need to be done to the array index.  I guess the answer is that the low bit is used as a flag?  

If I understand the issue properly this fixes a crash so it is much better than what was there.  However, it still isn't producing the right values.  If we can figure out how to get the right values we can improve the random encounters for temple+ by using a new encounter bitmap which we would need to make.  Sounds like a cool project for a future version.

doug1234:
BTW:  Should this fix issue #233?

DudeMcDude:
I'm not sure abiut the bitshift. I think i just copied the decompilation as is.

It would be nice to get proper x,y. You shouldn't have to generate a new bitmap, a proper solution should just scale it right. But i agree it's not critical for this release, the above is good enough I guess.

doug1234:
Part of the issue with the bitmap calculation is using integer math:  "bitmapW * (y * bitmapH / 565)" is not going to give an accurate result.  We can try changing the calculation to use only doubles and convert to int at the last step.  That will fix at least some of the problem.

doug1234:
I'll look into this one some more for next version.  But the crash is fixed so this is good for now.

Issue #372, opened 2020-01-09
Blackguard feat and Race model issues


vexagar:
Latest Co8 & Temple+ (1.0.75)

I couldn't get my Blackguard to select Divine Might, despite meeting prerequisites (13 Str, PA, TU) it remained greyed and unselectable.

Also, the female Wild Elf model is indistinguishable from male Wild Elf in character creation.

doug1234:
I had meant to comment on this one earlier.  Sorry about that.

I think the blackguard is correct.  Turn undead is needed for the divine might feat but the blackguard class has rebuke undead which I don't think qualifies.

The wild elf model should be fixed in the next release.

vexagar:
Thanks for the Wild Elf model fix /bow

I'm not fussed if that's how you want to rule Blackguard, they have plenty
of options to raise damage its not like they won't miss one more but I did
think it was inconsistent with the other four Turn/Rebuke feats that work
both ways. I try to read the hypertext d20 SRD but I'm a games-gamer, I've
never seen a pnp book in my life, so all I could find about this particular
feat was at another but seemingly related site, which state the
requirements were turn *or* rebuke:

Divine Might( Deities and Demigods
<http://dnd.arkalseif.info/rulebooks/supplementals-30--7/deities-and-demigods--39/index.html>,
p. 50)

[Divine <http://dnd.arkalseif.info/feats/categories/divine/index.html>]

The deity can channel energy to increase its damage in combat.
Prerequisite

Power Attack
<http://dnd.arkalseif.info/feats/players-handbook-v35--6/power-attack--2208/index.html>
 (PH) , CHA 13, STR 13, Ability to *turn or rebuke* undead,
Benefit

The deity can spend one of its turn/rebuke undead attempts to add its
Charisma modifier to its weapon damage for a number of rounds equal to its
Charisma modifier. This is a supernatural ability.
Also appears in

   1. Defenders of the Faith: A Guidebook to Clerics and Paladins
   <http://dnd.arkalseif.info/feats/defenders-of-the-faith-a-guidebook-to-clerics-and-paladins--38/divine-might--3639/index.html>
   2. Complete Warrior
   <http://dnd.arkalseif.info/feats/complete-warrior--61/divine-might--661/index.html>
   3. Faiths & Pantheons
   <http://dnd.arkalseif.info/feats/faiths-pantheons--18/divine-might--663/index.html>


[...]

PR #392, opened 2020-05-09
Blackguards can now use Divine Might issue #372.

doug1234:
Now blackguard turn undead ability will count for qualifying for Divine Might and other abilities.


doug1234:
Going to go ahead and close this one out.  Should be OK to include in a hotfix version if one becomes necessary.

PR #393, opened 2020-05-14
Add Swift Ambusher and Daring Outlaw feats also Atari Bug 81 and 90 fix

doug1234:


doug1234:
Going to go ahead an merge this one in before I add some more stuff.

PR #394, opened 2020-05-21
Dispel fixes

doug1234:
Several dispel fixes:  
dispel magic now works for > 3rd level spells
break enchantment now can't effect the same target twice in one casting
break enchantment is limited to transmutation and enchantment
dispel alignment/element touch attack now works to dispel spells (no check now) and kill creatures as appropriate
the level check cap is now enforced for dispel magic, break enchantment and greater dispel magic


doug1234:
Going to go ahead an merge this one in before I add some more stuff.

anatoliy-savchak:
Wow, awesome!

PR #395, opened 2020-06-06
PyObjHandle.conditions_get(int: kind = 0)

anatoliy-savchak:
This change is necessary for number of reasons, but mostly for Inspect functionality.


doug1234:
Looks good!  Merging in.

PR #408, opened 2020-07-18
Changed maxHpForNpcHitdice to HpForNPCHd

doug1234:
Changed maxHpForNpcHitdice to HpForNPCHd which can now be set to several options.  The old parameter will still work.  Added functions to override some configuration items.


doug1234:
One thing worth discussing would be which configuration items to expose through python.  Also, this average works a bit differently than the one for PCs which randomly uses either floor or ceiling when rounding off.  This one rounds up as is D&D standard.

Psionics-ToEE:
Could the option be added back into the game to allow you to select downed opponents as attack targets? I was getting some really annoying restrictions regarding this when I last played through the game, where it wouldn't let me perform attacks when clicking on opponents sometimes.

DudeMcDude:
> Could the option be added back into the game to allow you to select downed opponents as attack targets? I was getting some really annoying restrictions regarding this when I last played through the game, where it wouldn't let me perform attacks when clicking on opponents sometimes.

You can attack downed opponents, but you have to explicitly select attack from the radial menu, not as part of the default action click. It often happens that you click by accident and waste your attack.
That said, I guess it should be an option.

doug1234:
Since all comments have been good I am going to go ahead and merge it so anatoliy can test this out in his module.

PR #412, opened 2020-08-01
PerformBreakFree update and Inflict touch fix

anatoliy-savchak:
* PerformBreakFree updated with better code; tested;
* Inflict spell fixed to use melee, not ranged touch attack; tested;

### DIFF
--- a/TemplePlus/d20.cpp
+++ b/TemplePlus/d20.cpp
@@ -1174,8 +1174,21 @@ ActionErrorCode D20ActionCallbacks::PerformUseItem(D20Actn* d20a){
 }
 
 ActionErrorCode D20ActionCallbacks::PerformBreakFree(D20Actn* d20a) {
-	d20Sys.d20SendSignal(d20a->d20APerformer, D20DispatcherKey::DK_SIG_BreakFree, d20a->data1, 0);
-	return ActionErrorCode::AEC_OK;
+	int spellId = d20a->data1;
+	if (spellId) {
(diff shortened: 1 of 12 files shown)


doug1234:
A few comments:
(1)  Are you sure the inflict/cure spells being ranged touch attacks were a mistake and not written that way to work around other issues.
(2)  Seems like heal/harm could use a similar change...
(3)  It would be nice if they really worked correctly (so you could hold a charge).  Maybe a future fix?

anatoliy-savchak:
1) I suspect melee argument of touch function was added later. But currently it is obviously a mistake.
2) I agree, will alter, test and supply commit. Good point!
3) Yes I agree, that it would be nice to have. But I consider it rather niche functionality. Perhaps later.

DudeMcDude:
> 1. I suspect melee argument of touch function was added later. But currently it is obviously a mistake.
Yes, originally it was always ranged touch attack. I think for Searing Light. I added the flag for the vampiric touch fix.

anatoliy-savchak:
@doug1234 can you approve it? I think everyone has agreed on the changes.

doug1234:
Sorry about that.  I thought you were going to add the code you showed us on co8.  Merging now.

anatoliy-savchak:
@doug1234 - I did included counter* API and Inflict fixes into the Pull Request :)
The only thing I did not do is Harm, which I moved down in my backlog, as it is not critical at this moment.

PR #416, opened 2020-08-22
Fix invalidated iterator crash.

doug1234:
Fixes a crash that would happen from time to time closing windows.  The easiest way to get it was to bring up the map and then press escape (this was 100%).  A hide window message was causing the iterators to be invalidated when a window was removed from the list.  To fix it I went with a index based for loop.


DudeMcDude:
> Let me know if you see a better way to fix the issue...

I recall I had this problem with Object Events iteration...  take a look there.

DudeMcDude:
> > Let me know if you see a better way to fix the issue...
> 
> I recall I had this problem with Object Events iteration... take a look there.

Ok scratch that, it's irrelevant here 😞 
Your solution looks ok. I did find it odd that the townmap ui misbehaved and others didn't,  looks like it's because it doesn’t return 1 when intercepting the keystate message. I'll replace it just so it's in line with the others.

DudeMcDude (commented):
That's pretty ugly... how about adding a method to PySpell? spell.roll_dice_metamagic

doug1234:
Great idea.  I am going to fix false life and some other spells with similar problems but those fixes need to be on the C++ side.  So I will add something like that for C++ and python.

DudeMcDude:
> > > Let me know if you see a better way to fix the issue...
> > 
> > 
> > I recall I had this problem with Object Events iteration... take a look there.
> 
> Ok scratch that, it's irrelevant here 😞
> Your solution looks ok. I did find it odd that the townmap ui misbehaved and others didn't, looks like it's because it doesn’t return 1 when intercepting the keystate message. I'll replace it just so it's in line with the others.

Followup on this: looks like this was intended, to allow switching between UIs with hotkeys. I'm still a bit unclear how, but it generates an ESC keystate event when you press e.g. 'L' while in another UI.

PR #417, opened 2020-08-29
Fixed game.get_obj_by_id return null

anatoliy-savchak:
Fixed issue when string was referencing not existing id in Python.


doug1234:
Looks good to me.  Will merge after the weekend if there are no other issues found.  

What was the root issue that this fixed (did TOEE sometimes crash?) and will you need the change for your module?

anatoliy-savchak:
That was only related to Shattered Gates module. On game load code checked if stored object exist, and it was failing.

That's my bug, I wrote this code.

PR #418, opened 2020-09-04
UiTownmapMainWndMsg handler replacement to fix crash issue and bring …

DudeMcDude:
…inline with other UI msg handlers


doug1234:
So it will return TRUE now in response to the escape key.  This should fix the iterator that was being invalidated by breaking out of the loop.  Do you think I should keep my change to the loop on the other pull request or should I revert that part?

DudeMcDude:
Well, it's a fundamental issue, and I guess it could be encountered elsewhere, so keep it.

PR #419, opened 2020-09-10
Modified grease  to a 5 foot radius

doug1234:


DudeMcDude:
I know this is how it should be by the rules, but I have no doubt players would be pissed off by this. Let's leave it be.

doug1234:
How about a config option?  I hate to leave an obvious bug like this and I am pretty sure some will like it.

DudeMcDude:
Good idea. Let's call it "Stricter rules enforcement" for that sort of thing.
On that note, having a donut shape for reach weapons fits that as well.

doug1234:
Sounds like we are on the same page.  A donut shape for reach weapons is high on my list of things I would like to see.  I'll add the "Stricter rules enforcement" option shortly.

doug1234:
For donut correcting reach weapons, what would need to be done?  Would modifying CanMeleeTargetAtLocRegardItem, CanMeleeTargetFromLocRegardItem and CanMeleeTargetRegardWeapon be enough?

PR #427, opened 2020-09-23
Fix trap damage initiating combat

DudeMcDude:
Fixes issue reported here:

https://rpgcodex.net/forums/threads/temple-support-archive.99154/page-50#post-6684491

There were actually two things here:
One is that sprung traps force combat mode. This is a general (vanilla) issue.
The other is Co8 scripting that made Bassanio interrupt opening the chest. The problem was this occurs even when he's dead (but only once).

Also fixed a bug I introduced in the Module Switch branch that screwed up trap init.


doug1234:
Wow, another bug fix!  If you keep going at this rate, TOEE is going to be bug free! :)

DudeMcDude:
> Wow, another bug fix! If you keep going at this rate, TOEE is going to be bug free! :)

It's the annual harvest of the support thread :)

PR #431, opened 2020-09-28
Changed Craft MAA so that it lets you pick effects that you can't aff…

DudeMcDude:
…ord (so that you can see the required resources)


doug1234:
You are on a roll!!  The changes look good.  What was the effect of the world map bug?

DudeMcDude:
There was a small patch for the Co8 dll that prevented writing to unbound memory in the UiResize func. It was being applied to vanilla too, which caused a button to not be relocated and hence displaced.
Also I was using IsCo8 before ModSupport was inited.

PR #439, opened 2020-10-24
Fixes to Magic Circle

doug1234:
Now using spell names from the .mes file (I did it the easy way and only exposed the spell info)
Checking summoned creature for weapons and checking spell resistance (had to add a force since the spell does not allows SR).  Also checking that attacker is not a null object.
Did not switch to checking if spells are mind effecting.  Still using the list since it should not provide immunity to sleep, confusion and others.  May update protection from alignment later to match.
Update extended spell descriptions


DudeMcDude:
Awesome!
I suggest we freeze until 1.0.83 is out. Let's also prompt the testers to test this one.
Come to think of it, this will affect NPCs too, right? What happens when two circles of the same alignment clash?

doug1234:
Freeze sounds good to me.  I can let the testers know to try out the magic circle vs alignment spells.  If two circles of the same type overlap, protection from alignment effects will only be added the first time.  The second one will try but the effects will be rejected.  I tested that at one point...

PR #443, opened 2020-11-16
Easy new encounters

IPDTFE:
Hello, I'm back after all this time. I started a new branch because I was having issues with the older one. 

It is the same idea as the other currently open pull request, users can now create additional road encounters and sleep encounters by adding a file in a folder called encounterpacks located in their main TOEE directory. The folder's structure is fairly simple, I've attached an example.
[encounterpacks.zip](https://github.com/GrognardsFromHell/TemplePlus/files/5550039/encounterpacks.zip)

The program works by calling expanded_encounters (my script) instead of random_encounter in  python_integration_encounter.cpp.

expanded_encounters mimicks pretty much all of random_encounter's co8 version, limitations included.

If the user doesn't have a directory called encounterpacks in the TOEE folder or if co8 is not installed, random_encounter is immediately called by my script, ensuring normal functionalities.

Sleep encounters can be only called on one type of unit, but other units can be added through the reinforcements method.

Travel encounters work with RE_entry as normal.

Currently doesn't support linux I guess, haven't looked into proper path handling so it will behave as if there were no encounterpacks folder.

It should now fit most of the criteria that were set for me after the last upload, let me know if anything needs to change again.

As far as I've tested it works on vanilla and Co8 but my vanilla installation had a lot of separate issues for some reason so it might be worth rechecking on a different computer.

### DIFF
--- a/TemplePlus/python/python_integration_encounter.cpp
+++ b/TemplePlus/python/python_integration_encounter.cpp
@@ -201,8 +201,8 @@ static void __cdecl RandomEncounterCreate(RandomEncounter* encounter) {
 	if (encounterObj->map != -1) {
 		addresses.MapChange(); // No idea what this does
 	}
-
 	auto result = pythonObjIntegration.ExecuteScript("random_encounter", "encounter_create", args);
+
 	Py_DECREF(args);
 
 	if (!result) {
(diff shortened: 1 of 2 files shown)


DudeMcDude:
Welcome back!

TBH if you're just going to outright duplicate the Co8 random encounters file, and assume in your script it's a Co8 installation, you may as well make it a Co8 mod (perhaps an override file).
Otherwise, for this to be on the main branch, I want it to be as light and generic as possible.

Perhaps it would be better to run the original script first, and do a check on the encounter ID.
IDs in the range 2000-4000 indicate special stuff so hold off in that case, and if ID is in the 1000 range you can replace the RE.

PR #446, opened 2020-11-26
Craft Arms & Armor Update

doug1234:
Added new column for fixed extra price to arms and armor.  Improved crafted items long descriptions.  Added energy resistance armor and shields.


TemplePlus/ui/ui_item_creation.cpp
```diff
@@ -587,6 +589,10 @@ bool UiItemCreation::IsOutmoded(int effIdx){
 		for (auto it: appliedBonusIndices){
 			if (it == itEnh->upgradesTo)
 				return true;
+
+			//Removes the start effect after it is added (otherwise it displays the greyed start effect and the second effect)
+			if ((it == effIdx) && (itEnh->downgradesTo == CRAFT_EFFECT_INVALID))
```
DudeMcDude, line 594: Is this well tested?

TemplePlus/ui/ui_popup.cpp
```diff
@@ -3,27 +3,26 @@
 #include "ui.h"
 #include "tig/tig_keyboard.h"
 
-UiPopup uiPopup;
+UiPopupHandler uiPopupHandler;
```
DudeMcDude, line 6: I think I've unified this with UiPopup in my branch

TemplePlus/ui/ui_item_creation.cpp
doug1234: Define "well tested"...  it was definitely displaying strange as it was.  I made the change and it no longer displayed the greyed out first effect for spell resistance armor and the new energy armor.  I tested adding and removing effects and all looked good.  Is there anything else I should have tried?

TemplePlus/ui/ui_popup.cpp
doug1234: Right.  I changed a few things around in addition to renaming.  I see a messy merge in my future...

TemplePlus/ui/ui_item_creation.cpp
DudeMcDude: Does it just affect display? What happens if you take an already crafted item and then remove an effect?
I just have a vague feeling of deja vu here...

doug1234: It is display only but odd enough I thought I had added a bug at first.  I am pretty sure I tested adding then backing out effects and everything was good.  I'll give it another test after my merge though.

DudeMcDude: Just to be clear, I'm talking about the case where you hit the craft button and reopen the craft UI to add more stuff. You need to make sure that doesn't go haywire.

doug1234: I can add an effect, craft the item and then come back, add the next effect and craft again.  If that is what you mean then it works.  I have seen another graphical oddity where effects show up double on the right side but that is a different (display only) issue I'll look into at another time.  Merged and crafting still seems to be working correctly.

PR #453, opened 2020-12-05
Added Agility and Stamina armor.

doug1234:
Added support for and as well as or conditions.  Not enough to support really weird stuff but probably everything we will add.


doug1234:
This pull request is getting kinda long.  I am going to merge it fairly soon if no issues are found.

DudeMcDude:
Can you verify that nothing ingame is directly referencing the shield condition?

doug1234:
Good thought.  I can see why that would be a concern.  

I only see condShieldBonus referenced in the itemConditions array in the decompile.  I don't see any reference in Temple+.  Anything else I should look for?

DudeMcDude:
Yup, it isn't. 
I think there was some xomplaint about armor radial effects being under different armor submenus?

doug1234:
I thought about putting that one in here but I decided only to add the .mes and .tab changes and leave it disabled.  There was indeed an issue with the armor menu but the bigger thing was that you could take off the armor to refresh the charges.  Not sure how to deal with that.  I think this pull request is big enough and I'll make a pull request for the blurring and displacement armor if I figure out how to deal with the charges incorrectly refreshing...

doug1234:
I am pretty convinced this is safe to merge.  I want to get this in for testing and add more stuff in another pull request.

TemplePlus/condition.cpp
```diff
@@ -5352,6 +5375,44 @@ int ItemCallbacks::BucklerToHitPenalty(DispatcherCallbackArgs args)
...
+int __cdecl ItemCallbacks::MaxDexBonus(objHndl armor)
+{
+	auto res = itemCallbacks.oldMaxDexBonus(armor);
+	
+	//Query for max dex bonus adjustment
+	if (armor) {
+		auto parent = inventory.GetParent(armor);
+		if (parent) {
+			auto obj = objSystem->GetObject(parent);
+			if ((obj != nullptr) && (obj->GetInt32(obj_f_type) == obj_t_pc)) {
```
anatoliy-savchak, line 5387: Why PC only? How about NPC?

DudeMcDude: Why indeed.
Also please use obj methods like IsPc, IsNpc, IsCritter in case we ever refractor this.

doug1234: Both valid suggestions. I'll update this in my next pull request.

PR #455, opened 2021-01-12
Added Silent Moves and Shadow to craft options

doug1234:


DudeMcDude:
Don't forget to add to changelog!

doug1234:
Change log updated.

DudeMcDude:
Are you planning on adding anything more? Otherwise I want to release this weekend.

doug1234:
Just merged it in.  I have some more stuff I will do related to crafting but that can be in another version.  Do you think this version has been tested enough?  If so, I could do the release on Friday if you want.

BTW:  I noticed that if you turn on the stacking metamagic, extra empower does not seem to have an effect.  Is that OK for this version?

DudeMcDude:
> Just merged it in. I have some more stuff I will do related to crafting but that can be in another version. Do you think this version has been tested enough? If so, I could do the release on Friday if you want.
> 
> BTW: I noticed that if you turn on the stacking metamagic, extra empower does not seem to have an effect. Is that OK for this version?

Yeah, no biggie.
Friday it is!

PR #457, opened 2021-01-16
savehook changes to support new modules using save archivation

DudeMcDude:
As requested by @anatoliy-savchak

### DIFF
--- a/TemplePlus/gamesystems/gamelib_load.cpp
+++ b/TemplePlus/gamesystems/gamelib_load.cpp
@@ -186,6 +186,13 @@ bool GameSystems::LoadGame(const string& filename) {
 		return false;
 	}
 
+	{
+		// user savehook
+		auto saveHookArgs = Py_BuildValue("(s)", filename.c_str());
+		pythonObjIntegration.ExecuteScript("templeplus.savehook", "load", saveHookArgs);
+		Py_DECREF(saveHookArgs);
+	}
+
 	addresses.UiMmRelated2(63);
 	
 	logger->info("Completed loading of save game");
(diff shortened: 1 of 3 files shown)


TemplePlus/gamesystems/gamelib_save.cpp
```diff
@@ -211,6 +211,14 @@ bool GameSystems::SaveGame(const string& filename, const string& displayName) {
...
 	tio_fclose(file);
 
+	{
+		// user savehook
+		auto saveHookArgs = Py_BuildValue("(s)", filename.c_str());
+		pythonObjIntegration.ExecuteScript("templeplus.savehook", "save", saveHookArgs);
+		Py_DECREF(saveHookArgs);
+	}
+
+
```
shartte, line 221: What do you think about moving this down to directly before `// Create savegame archive`?

shartte: Also: I'd probably create an absolute path for `"Save\\Current"` (which is passed to SaveGameArchive), and pass that to the hook so it can add its own files to the archive more easily.

TemplePlus/gamesystems/gamelib_load.cpp
```diff
@@ -186,6 +186,13 @@ bool GameSystems::LoadGame(const string& filename) {
 		return false;
 	}
 
+	{
+		// user savehook
+		auto saveHookArgs = Py_BuildValue("(s)", filename.c_str());
```
shartte, line 191: Instead of filename, I'd pass the absolute path to `"Save\\Current"`

anatoliy-savchak:
@DudeMcDude, when this pull request will be merged?

PR #471, opened 2021-02-09
Fix EK_Q_Critter_Has_Spell_Active handler in new spells.  Fix #470.

### LINKED ISSUE #470: New spells have buggy XSpellIsActive callbacks
Example:
https://github.com/GrognardsFromHell/TemplePlus/blob/183509ba001ac320932883df702d3250b7c48df0/tpdatasrc/tpgamefiles/scr/tpModifiers/sp_moment_of_prescience.py#L108

This will return 1 regardless of the spell enum being queried. Looks like a forgotten placeholder.

doug1234:


DudeMcDude (commented):
This is a good start. But, notice that the original HasSpellActive callback also returns the cond struct. Not sure how critical that is, but it's best not to break it. 
This also begs to be implemented as a generic template.

doug1234:
A generic implementation sounds reasonable.  Not sure on returning the cond struct though.  Is there a way to do that from python or would that require a C++ implementation?

DudeMcDude:
I think it's better to create a standard handler in C++.

doug1234:
OK.  Works for me.  I think I could create a generic handler in python pretty easily but that can be for another time.

PR #474, opened 2021-02-23
Swapped a couple feats to a swift action, Added precise weapon and added disable sneak attack range requirement python query

doug1234:


DudeMcDude (commented):
I think Disable AOO already exists elsewhere, but from the attacker’s perspective. Have you checked that?

doug1234:
> 
> 
> I think Disable AOO already exists elsewhere, but from the attacker’s perspective. Have you checked that?

Yes.  As you said, it is from the attacker's perspective.  Not sure how I would use that to disable the AOO when the target has the feat and a shield.  However, the names are a bit similar.  Do you think I should change it to something like "Don't Provoke AOO"?

BTW:  Looks like the Visual Studio on the build computer doesn't like the dependencies in my latest CI build.  Is it time to rebuild the dependencies package?

DudeMcDude:
Come to think of it, I think there's already a query for that? Something like aoo incurs?

doug1234:
Interesting idea... There is AOOIncurs but it doesn't take the reason for the AOO as a parameter its parameters.  It takes the attacker handle as a parameter.  I don't know if that will work since I only want to cancel spell casting AOOs only.

doug1234:
Should I merge this one or do you have another idea on avoiding the AOO?

DudeMcDude:
Wait, what about the ActionTriggersAoo query just below that? Is it because the action you're after doesn't have that flag?

doug1234:
Cast defensively sets the D20ADF::D20ADF_QueryForAoO flag to 0 (I think) so I am not sure that would do the trick.  Let me give that a shot and see if there is a way to make it work.

DudeMcDude:
> Cast defensively sets the D20ADF::D20ADF_QueryForAoO flag to 0 (I think) so I am not sure that would do the trick. Let me give that a shot and see if there is a way to make it work.

It sets the query return value to 0, not the flag.

doug1234:
Interesting.  I'll take another look at it tomorrow night.

doug1234:
I made the change you suggested.  That was indeed a better way to do that.

PR #496, opened 2021-05-09
Enervation/Energy Drain

dolio:
This is an attempt to put together all the resources for Enervation and Energy Drain.

The spells already have numbers allocated in the Troika .mes files, because they're core spells. However, obviously a lot of stuff was left out. So far I've added the spell scripts and rules text files. I have other things written, but I'm unsure where to put them. Here's a breakdown of everything I have somewhere:

1. Either the Troika or Co8 install has lines in `data/snd_spells.mes` for Enervation. But those lines are missing for Energy Drain. I think it'd be fine to just duplicate the lines, but there's no `snd_spells.mes` in TemplePlus that I can see. My guess was `tpdata/tpmes/sounds.mes`, and editing that in my ToEE directory works, but I wanted to check first.
2. I was able to make help entries in an overrides extension file. However, when I tried moving those lines to `tpdata/tpmes/help_extensions.tab` in my ToEE directory, it didn't seem to work. Should it go in the `co8infra` `help_extensions.tab` instead? I can't easily pack that up myself to check if that will work.
3. It seems like only `tpgamefiles` has a `spell_long_descriptions.mes` file. Is that the one that's correct to modify?

I think that's what I have outstanding, and it seems like most functionality is working appropriately. If I've overlooked some addition let me know, though.

One thing I've left out so far is trying to detect if these spells are going to kill the target, and dealing a little damage so that experience is rewarded. That's because I can't reliably tell in python whether the target will actually be killed (because there's no way to access the corrected hit dice num function). I think that's probably not a huge issue, because these are mainly debuffs. I might see if I can find a solution, though.


doug1234:
Wow.  Great stuff!

1 - tpgamefiles\sound\user_sounds\tpsounds.mes should be updated for sound.
2 - Update the co8infra one and probably the kotbfixes one also
3 - tpgamefiles\mes\spell_long_descriptions.mes only exists in one place where it should be modified

dolio:
One additional thing I found: the short descriptions (? entries 5150 and 5151) in spell.mes are there in the Troika file, but bad. They just say `Energy Drain` and `Enervation` respectively. Should that go in the `spell_ext.mes`? I'm not actually sure where those even show up in the game.

DudeMcDude:
2. tpdata is added to path relative to the Temple+ executable, rather than the ToEE folder.

About xp: the engine attributes the kill to obj_f_who_hit_me_last. In Temple+ you can finally set this via the new PyObjHandle method .obj_set_obj :)

dolio:
Oh, great! That's much easier.

DudeMcDude:
> One additional thing I found: the short descriptions (? entries 5150 and 5151) in spell.mes are there in the Troika file, but bad. They just say `Energy Drain` and `Enervation` respectively. Should that go in the `spell_ext.mes`? I'm not actually sure where those even show up in the game.

Yeah, spell_ext.mes.
Those are the descriptions shown in the Char/Char Editor UI description box when you hover the spell.

DudeMcDude:
> Oh, great! That's much easier.
> 
> Edit: Temple+ is in my ToEE folder, incidentally. I was just modifying files in place to test things out.

Hmmm, was that with the dev build or release build? The latter looks for ../tpdata I think.

dolio:
It was with one of the nightlies. I fooled around some more with it and it still didn't show up in the help.

I can add it to the tab files and see if it shows up using the artifacts from the PR build.

DudeMcDude:
> It was with one of the nightlies. I fooled around some more with it and it still didn't show up in the help.
> 
> I can add it to the tab files and see if it shows up using the artifacts from the PR build. Or will that not work, because they're not packed into the .dat files automatically?

Oh right, I forgot that the file is also included in co8infra.dat. This has higher priority than the tpdata folder. So yeah, packing it should work. It's not done automatically however.

DudeMcDude:
So is there anything else or is it ready?

DudeMcDude:
That's ok, us windows plebes can take care of that ;)

DudeMcDude:
Now packed in the dats. Congrats!

PR #498, opened 2021-05-14
Fix Camera Issues in AdvanceTime/ProcessMessages when renderSize!=screenSize

### LINKED ISSUE #462: [Linux/Wine] Special effects (particles?) are visible only in a portion of the screen.
Hello, it's me again.
Don't mean to spam the bugtracker, just reporting bugs as I go (and this seems the place with any activity for T+).

This particular one is very frustrating...
I've found that various special effects (spells such as sleep/grease/MM/web/etc, chimney smoke, blood, will-o-wisps, bugs(?) around trees in Nulb, even flames in the main menu) are visible (get rendered?) only in a portion of the screen.

Now, I have a 4k monitor running at 3840x2160, and have render{width,height} set to eg 1280x720 fullscreen. And all the special FX would only show up in a small rectangle (or a square) in the lower right corner of the screen.

I think T+ doesn't change the physical resolution to render{width,height}, but rather scales to whatever the desktop res is? At least it is so with Wine. Probably a smart thing to do (however, the fonts get blurry, they have to be rendered at the actual resolution).

All three w-o-w's are visible:
3wow

After scrolling the screen, only one (the rightmost) is visible:
2wow

Scrolling the screen even further, none are visible:
0wow

Also notice the green glow from the Bless spell around the chars (not very pronounced on the screenshot) behaves the same. 

If I set the resolution to say 1920x1080 this area gets bigger.

I've set debugPartSys=true, but all I see is that the blue rectangles with labels like sp_Bless (which is I presume for the green glow) or sp_Sleep (for the corresponding spell) are somewhere off-screen in a diagonally opposite corner to which has the effects actually visible:
spdebug

Thanks again for helping to get T+ even working for me in Wine.

shartte:
Reduce reliance on RenderingDevice::GetCamera() since it can switch projection when the render target changes. This causes issues if any code in ProcessMessage/AdvanceTime makes use of the camera to determine visibility of objects (i.e. particle systems). This happens because the gameView render target is only pushed during RenderFrame and when it is popped after rendering is done, the RenderingDevice will automatically call SetScreenSize on the camera using the backbuffer's size.

Instead, use GameView (global 'gameView') to keep a camera that always represents the view/projection of the actual game view, which is never touched.

This should fix #462.

### DIFF
--- a/Infrastructure/include/graphics/camera.h
+++ b/Infrastructure/include/graphics/camera.h
@@ -1,6 +1,8 @@
 
 #pragma once
 
+#include <memory>
+
 #include "../infrastructure/location.h"
 #include "collision.h"
 #include "math.h"
@@ -12,7 +14,6 @@ namespace gfx {
 	 */
 	class WorldCamera {
 	public:
-
 		bool IsBoxOnScreen(XMFLOAT2 point, 
 			float left, float top, float right, float bottom) const;
 		
@@ -48,7 +49,10 @@ namespace gfx {
 			return mCurScreenOffset;
 		}
 
-		void SetScreenWidth(float width, float height) {
+		void SetScreenSize(float width, float height) {
+			if (std::abs(width - mScreenWidth) < 0.01f || std::abs(height - mScreenHeight) < 0.01f) {
+				return;
+			}
 			mScreenWidth = width;
 			mScreenHeight = height;
 			mDirty = true;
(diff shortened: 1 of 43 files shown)


shartte:
@DudeMcDude This needs some testing and is not as nice as I'd like it to be, but it should fix the issue (and probably others)

PR #501, opened 2021-05-18
Factor and add to spell_utils.py

dolio:
- Split out some functions for getting names from a spell id
- Add a function for formatting spell durations nicely
- Add several functions and a subclass for implementing spells
  that use the 'holding charge' mechanics.


DudeMcDude (approved):
I'm going to merge it as is, but may make some changes so stay tuned

PR #504, opened 2021-05-29
Abjurant champion

doug1234:
Draft of the abjurant champion prestige class.

The one issue I have is that the arcane boost action brings up the picker.  The only thing that can be selected is the character that is preforming the action (which is what I want).  Still it would be better if the picker was not displayed at all.


tpdatasrc/tpgamefiles/rules/d20_actions/action02900_abjurant_champion_arcane_boost.py
```diff
@@ -0,0 +1,22 @@
...
+import tpactions
+
+def GetActionName():
+	return "Arcane Boost"
+
+def GetActionDefinitionFlags():
+	return D20ADF_None
+	
+def GetTargetingClassification():
+	return D20TC_CastSpell
```
DudeMcDude, line 11: So why do you actually need this to be d20tc_castspell?

doug1234: I need the spell packet for getting the spell level and debiting the spell.  If I do D20TC_Target0 (which would not show targeting), I don't get a spell packet.

DudeMcDude: Well that data is set according to what's already stored in evt_obj.d20a.spell_data, so you could just retrace it by creating a tpdp.SpellPacket(), setting its fields just like D20::TargetCheck() does and then calling debit from it. If there isn't already you should add a helper method for creating a SpellPacket from D20SpellData.

doug1234: Thanks!  That is an interesting idea.  Looks like a spell packet can currently only be created from a spell id but I'll see about making a way to construct one from D20SpellData.

doug1234: That seemed to work great.  I pushed up the change.  I'll test just a bit more and remove the draft tag on Friday.

DudeMcDude:
Awesome, so is it ready for merge?

doug1234:
It is done but I wanted to test it a bit more tomorrow night.  Then I will remove the draft tag.

doug1234:
Should be ready now.

PR #505, opened 2021-05-29
Spell utils

DudeMcDude:


tpdatasrc/tpgamefiles/scr/spell_utils.py
```diff
@@ -0,0 +1,275 @@
...
+#[pytonModifier].AddHook(ET_OnConditionAdd, EK_NONE, spell_utils.addDimiss, ())
+def addDimiss(attachee, args, evt_obj):
+    spellPacket = tpdp.SpellPacket(args.get_arg(0))
+    spellPacket.caster.condition_add_with_args('Dismiss', args.get_arg(0))
+    return 0
+
+#[pytonModifier].AddHook(ET_OnD20Signal, EK_S_Temporary_Hit_Points_Removed, spell_utils.removeTempHp, ())
+# needed in combination with condition_add_with_args('Temporary_Hit_Points', 0, spell.duration, tempHpAmount)
+def removeTempHp(attachee, args, evt_obj):
+    attachee.d20_send_signal(S_Spell_End, 'Temporary_Hit_Points')
```
DudeMcDude, line 88: This is unspecific to any particular spell... could cause issues if multiple spells add temp HP this way.
@Sagenlicht

tpdatasrc/tpgamefiles/scr/spell_utils.py
```diff
@@ -0,0 +1,275 @@
...
+
+#[pytonModifier].AddHook(ET_OnD20Signal, EK_S_Killed, spell_utils.spellKilled, ())
+def spellKilled(attachee, args, evt_obj):
+    args.remove_spell()
+    args.remove_spell_mod()
+    return 0
+
+#[pytonModifier].AddHook(ET_OnD20Signal, EK_S_Spell_End, spell_utils.spellEnd, ())
+def spellEnd(attachee, args, evt_obj):
+    print "{} SpellEnd".format(spellName(args.get_arg(0)))
```
DudeMcDude, line 62: Shouldn't spellEnd actually do something? The common spell ender does various things.
@Sagenlicht

tpdatasrc/tpgamefiles/scr/spell_utils.py
Sagenlicht: This should only be triggered if Temp HP value <= 0, no matter from what sources Temp HP were originally granted. My bad, if I did not make this clear.

I just noticed, when I use attachee.condition_add_with_args('Temporary_Hit_Points', 0,  duration, tempHP) the condition does not automatically gets removed when temp HP reach 0 but requires an spell_end signal.

When TempHP falls to or below 0, a signal is triggered (S_Temporary_Hit_Points_Removed) but it actually requires the spell triggering then the removal of the Temporary_Hit_Points condition by doing the added line in spell_utils.

It basically looks like:
def righteousFurySpellOnConditionAdd(attachee, args, evt_obj):
    attachee.condition_add_with_args('Temporary_Hit_Points', 0, 600, args.get_arg(2)) #temp HP value is passed by spell (arg3), duration is up to 1 hour;
    return 0
def righteousFurySpellRemoveTempHp(attachee, args, evt_obj):
    attachee.d20_send_signal(S_Spell_End, 'Temporary_Hit_Points')
    return 0
righteousFurySpell.AddHook(ET_OnConditionAdd, EK_NONE, righteousFurySpellOnConditionAdd,())
righteousFurySpell.AddHook(ET_OnD20Signal, EK_S_Temporary_Hit_Points_Removed, righteousFurySpellRemoveTempHp, ())

I now want to replace the signal handling by:
righteousFurySpell.AddHook(ET_OnD20Signal, EK_S_Temporary_Hit_Points_Removed, spell_utils.removeTempHp, ())

tpdatasrc/tpgamefiles/scr/spell_utils.py
Sagenlicht: This is just an additional hook that prints to the console (the end of the spell). I just adopted it from spells like Moment of Prescience. I rarely do anything on Spell End in my spells, e.g. spell particles that are permanent are not cleared by my spells but the game ends them. I always assumed (and still do!) that the game ends all particles registered within a spell_id when the spell ends.
Long story short, Close to all my spells have this exact hook with just a print to the console. If this is not required, I think we can just drop the hook and only use it, when it is actually required to do something in spellEnd.
An example where I actually do something in spellEnd would be Iron Guts that adds a condition once the spell expires.

DudeMcDude: Right, I mixed this up with remove_spell which has a default handler, which together with the generic AoE handler is probably good enough for 99% of cases.

And yeah, if all it does is print to console, it can be safely removed, since you'll get a log message from remove_spell anyway.

tpdatasrc/tpgamefiles/scr/spell_utils.py
DudeMcDude: Are you sure it's only triggered by that?
Also note that attache.d20_send_signal(S_Spell_End, 'blah blah') is meaningless, and is parsed internally to attache.d20_send_signal(S_Spell_End, 0, 0).
I believe that if you cast two different spells with this temp HP scripting, removing the temp HP of one (due to damage) will end the other.

tpdatasrc/tpgamefiles/scr/spell_utils.py
```diff
@@ -0,0 +1,275 @@
...
+    #
+    # the given argument number is for additional arguments beyond
+    # this.
+    #
+    # The standard countdown hook is not added, since the typical
+    # mechanic is for charges to be held indefinitely until used.
+    # An argument is reserved for time limited touch spells,
+    # however, and the countdown hook can be added in such a case.
+    def __init__(self, name, argn):
+        PythonModifier.__init__(self, name, 3+argn)
```
DudeMcDude, line 247: Something to keep in mind: by default this adds an 'Add Only Once' hook. What would happen if you cast the same spell twice?
@dolio

tpdatasrc/tpgamefiles/scr/spell_utils.py
Sagenlicht: No, I am not 100% sure, but I tested a lot with Temporary Hit Points and tried to understand what the game actually does. I use the vanilla condition 'Temporary_Hit_Points' (https://github.com/GrognardsFromHell/DllDocumentation/wiki/Conditions-(List)-page-2-M-Z#temporary_hit_points) for all my spells that grant Temp HP. Basically I add the condition as a secondary condition. I belive this exactly what the vanilla spell Aid does.

When the Temp HP are reduced to 0 or below the 'Temporary_Hit_Points' condition does not get removed (which I originally hoped it would) and I digged around in the C code to understand what the Aid spell does. 

Both the spell https://github.com/GrognardsFromHell/DllDocumentation/wiki/Conditions-(List)-Spells#sp-aid and 
https://github.com/GrognardsFromHell/DllDocumentation/wiki/Conditions-(List)-Spells-page-2#sp-false-life 
react to the DK_SIG_Temporary_Hit_Points_Removed signal.

I did not test what happens if more than one source grants temp HP, I did not test if the condition can be applied more than once, but I tested and know 100% that the signal will only remove the Temporary_Hit_Points condition and not my original spell condition. For example Righteous Fury grants + 4 Str and Temp HP, when the Teemp HP is gone the condition gets removed but the original spell condition stays and still grants +4 Str.

tpdatasrc/tpgamefiles/scr/spell_utils.py
Sagenlicht: Ok then let's remove it :)

tpdatasrc/tpgamefiles/scr/spell_utils.py
dolio: This was intentional (if you're talking about what I think you're talking about).

Casting a spell while you're holding charge causes you to lose the held charge. So there should never be a situation where you have two copies of the 'holding charge' condition for a spell.

tpdatasrc/tpgamefiles/scr/spell_utils.py
DudeMcDude: As you said, both Aid and False life react to that signal... by sending d20_send_signal(S_Spell_End, **spell.id**, 0). If they hadn't specified the spell ID, they'd also end each other's temp HP.

tpdatasrc/tpgamefiles/scr/spell_utils.py
DudeMcDude: I think we're talking about something different.

PythonModifier has an optional arg preventDup, which is true by default. This adds a generic hook for ET_OnConditionAddPre, which prevents the new condition from being added.

My main concern is that when the 2nd spell instance is cast it may produce an unterminated spell, since the spell cast goes through as normal but the condition isn't actually applied and so no duration countdown actually occurs.

The "preventDup" hook fires before the touchAdd (ET_OnConditionAdd) is called so the S_TouchAttackAdded signal will not fire, and consequently neither will touchTouchAttackAdded which removes the old charge.

In general, I think most spells do not have the preventDup generic hook, so spell modifiers should have this off by default as well.
Incidentally, I just checked the spell condition list which ones have it (0x100ECF30), and it looks like Call Lightning and Call Lightning Storm has this issue as well.

dolio: Okay, so should spell modifiers in general not use the generic hook, because they need a more specific hook that e.g. cancels the spell in addition to not adding the condition?

And in this case, the older spell should cancel itself via the other hooks I've added, so the new spell doesn't need to cancel itself or prevent its effect from being added?

dolio: So, I just tried out my spell that uses this, and it seems like it actually works correctly. When I cast the spell a second time, the original spell stops ticking, and the new one starts ticking (based on the `OnBeginRound` functions in the spell scripts).

I'm not sure how that's happening exactly. It seems to work the same when I tell it not to add the generic hook.

DudeMcDude: Oh, it's through the S_Spell_Cast signal which ends the previous spell whenever any spell is cast. This is invoked during the PerformCastSpell action so that's why it happens beforehand.
Although that brings up a different issue, I think this will also happen when any spell is cast on you as well (as a target, rather than caster, that is).

tpdatasrc/tpgamefiles/scr/spell_utils.py
Sagenlicht: Ok I think, I understood what you want me to do.
You basically want me to add the spell.id to the condition, Will do that and modifiy the spell_utils as well then.
Instead of:
`attachee.condition_add_with_args('Temporary_Hit_Points', 0, duration, tempHpAmount)`
it should be:
`attachee.condition_add_with_args('Temporary_Hit_Points', spell_id, duration, tempHpAmount)`

and then in the spell_utils:
`def removeTempHp(attachee, args, evt_obj):`
` attachee.d20_send_signal(S_Spell_End, 'Temporary_Hit_Points', args.get_arg(0))`
`    return 0`

Will do that.

Sagenlicht: After changing it to the above, the Condition does not get removed anymore when the signal is send :( 

Will need to spend some more time on this then.

DudeMcDude: `attachee.d20_send_signal(S_Spell_End,'Temporary_Hit_Points', args.get_arg(0))`
This is the problem part. Again, the string arg is meaningless and will be parsed by the engine to 0. Remove it and it should work.

Sagenlicht: Ok, wow guess brain afk, got it. Will fix it and ty.

DudeMcDude: I should probably make it throw an error in such cases :)

tpdatasrc/tpgamefiles/scr/spell_utils.py
dolio: Yeah, you're right. Casting a spell on the person holding the charge eliminates the charge. I think Touch of Fatigue has this same problem.

Is there a way to just detect when the subject of the condition casts a spell? Is it just a matter of inspecting the d20 action in the event object to see who is performing the spell casting?

DudeMcDude: Heh, apparently it's like that for most vanilla spells. Another one for the list :)

You should be able to compare attachee == spell_packet.caster in the S_Spell_Cast handler.
@doug1234 might want to fix that in Touch of Fatigue too.

doug1234: I'll take a look at that after I finish up that draft pull request.

PR #506, opened 2021-05-29
approach support in API

anatoliy-savchak:
* PyObjHandle.find_path_to_obj - need it to make NPC approach other NPC;
* PyObjHandle.standpoint_get - need it to check where NPC is supposed to be;
* disable WieldBest in ZMOD.

### DIFF
--- a/TemplePlus/condition.cpp
+++ b/TemplePlus/condition.cpp
@@ -5388,6 +5388,7 @@ int ItemCallbacks::UseableItemActionCheck(DispatcherCallbackArgs args){
 
 	if (objType == obj_t_scroll && !spellSys.CheckAbilityScoreReqForSpell(args.objHndCaller, spData.spellEnum, -1) && !useMagicDeviceSkillBase){
 		dispIo->returnVal = AEC_INVALID_ACTION;
+		logger->info("CheckAbilityScoreReqForSpell: Not enough ability score to use scroll!");
 		return IEC_Cannot_Wield_Magical;
 	}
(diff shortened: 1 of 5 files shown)


TemplePlus/condition.cpp
```diff
@@ -5388,6 +5388,7 @@ int ItemCallbacks::UseableItemActionCheck(DispatcherCallbackArgs args){
 
 	if (objType == obj_t_scroll && !spellSys.CheckAbilityScoreReqForSpell(args.objHndCaller, spData.spellEnum, -1) && !useMagicDeviceSkillBase){
 		dispIo->returnVal = AEC_INVALID_ACTION;
+		logger->info("CheckAbilityScoreReqForSpell: Not enough ability score to use scroll!");
```
DudeMcDude, line 5391: Does this produce logspam? (being checked constantly)

anatoliy-savchak: True. It will be checked many times for a scroll. But there is a problem of constant debugging of scrolls usage. Often it is unclear what was the problem of spell not firing. I had to add it. If it will live, I will submit probably more such logs. But not now, let's wait how it will present itself.

DudeMcDude: I'm not sure I understand what you wrote, but at any rate log spam is not acceptable.

For your purposes I can offer changing this log message to logger->trace and adding a config setting for the log level so that it is not normally visible. (I'll add this config)

PR #508, opened 2021-06-01
Spell Compendium Admin Files

Sagenlicht:
The second help tab (generell_help_modifier.tab)  contains a few lines that overwrite existing lines in help.tab. This is done to support the new spells help topic.


DudeMcDude:
> The second help tab (generell_help_modifier.tab) contains a few lines that overwrite existing lines in help.tab. This is done to support the new spells help topic.

Ok that’s an important issue actually. 
You have to remember that ToEE also has non english versions. With that in mind, we should override as few original entries as possible, at least in the general tpgamefiles.
You have more leeway in co8infra since that is only added when Co8 is installed, which is inherently english only.

The other thing is that I don't think it's necessary to add the new classes to the spell help files. It'll just be a huge mess, keeping it core-only is fine and the new classes can link to their own spell lists.

Sagenlicht:
Ok, should I completly drop the new spells overview entry in the help then?

These are the new entries I made:
image
Spells List scrolls down for all casting classes:
image
Example window for a spell level (where you can select the according spell for details)
image

Modifier entry:
image
Clicking on the entries describes the bonus/modifier type (flavor)

To make this work, I had to modify following help entries:
* All core assassin and blackgurad spell help entries (e.g. adding TAG_BLACKGUARD_1 as a tag to every core Blackguard level 1 spell)
* Modify the blackguard prestige class to support spell lists (you can now jump to the spell list now from the class entry)
* Modifiy the assasin prestige class to support the spell lists (you can now jump to the spell list now from the class entry)
* Modify Tag_Root (adding Modifier and Spell List Tab)

All actual spell help files for the new spells are stored in a different extender file, they are not influenced by that.

I would like to keep the Modifer entries, I linked to them in a lot of spell help entries, but they can be kept without a root entry if needed.

DudeMcDude:
You can add this stuff to co8infra.

BTW this discussion is relevant for mes files as well.

Sagenlicht:
> You can add this stuff to co8infra.
> 
> BTW this discussion is relevant for mes files as well.

I am unsure what mes files you a refering to, but I personally would have no problem to move all my stuff to co8 if this prefered? Can also do a new folder that reqiuires co8infra if this would be even better for you? 

In the end, I don't think, it makes a lot of sense to use new spells but ignore the co8 stuff anyways? I have never thought about this though, cause I use the co8 stuff since it was released.

DudeMcDude:
> > You can add this stuff to co8infra.
> > BTW this discussion is relevant for mes files as well.
> 
> I am unsure what mes files you a refering to, but I personally would have no problem to move all my stuff to co8 if this prefered? Can also do a new folder that reqiuires co8infra if this would be even better for you?
> 
> In the end, I don't think, it makes a lot of sense to use new spells but ignore the co8 stuff anyways? I have never thought about this though, cause I use the co8 stuff since it was released.

Just to make sure we're on the same page, I'm talking about the co8infra in /tpdatasrc, that is then packed to co8infra.dat.

Re. Mes files, for example the damage.mes file.

tpdatasrc/tpgamefiles/mes/damage.mes
```diff
@@ -0,0 +1,89 @@
...
+{2008}{~Adamantine~[TAG_ADAMANTINE]}
+{2009}{Bludgeoning}
+{2010}{Piercing}
+{2011}{Slashing}
+{2012}{~Mithril~[TAG_MITHRAL]}
+{2013}{Cold Iron}
+{2014}{--}
+
+// Added by Sagenlicht
+{3000}{~Critical Strike~[TAG_SPELLS_CRITICAL_STRIKE]}
```
DudeMcDude, line 82: Are you sure these even work? I think Temple+ looks up tpmes/damage.mes rather than mes/damage.mes

Sagenlicht: I don't know :( I am used to have the files in the ToEE  overrides folder and then I have to place them in overrides/mes/damage.mes. 

I wasn't aware Temple+ requires a different file structure, I am sorry about that. I didn't find the time to work on this today, I will push an update tomorrow then.

DudeMcDude: I added tpmes/damage.mes so as to not override the regular damage.mes. That was a long time ago however.
In general I think I'm going to have to bite the bullet and allow for composing mes files from different sources so I don't have to create extender handling for every single one.

Sagenlicht: I only need the modifications in damage.mes because I can't use a string in damagePacket add_dice.

It would make more sense to me, to actually support strings for that in tpdp. I accidentally thought you have done that. I personally would not need a modified damage.mes if custom strings would be supported.

A whole different story would be the radial.mes. I have to modify it for radial options in spells. We talked about that in the doc in length, you said you add it to the to do list.

DudeMcDude: Yeah that's what I mean. There's a ton of mes files and the list of extenders keeps growing. I want to solve this at the root.

Sagenlicht: Ok, just to be sure, we are on the same page.
I move 
> damage.mes 
spell_radial_menu_options.mes 
generell_help_mods.tab (contains the modifactions to help)

to co8infra/tpmes (in this folder is help_extensions.tab atm)

and keep the rest of the files in tpgamefiles? Which would be:
>buff and debuff symbols
spell_compendium_help.tab (contains the help to all new spells)
spell_compendium_spell.mes (short description) (though I am unsure if I need a subfolder in mes)
spell_compendium_spell_long_descriptions.mes (long descrption)
blackguard and assassin char_class files
spell_compendium_partsys.tab
spell_compendium_spell_enum.mes
spell_compendium_sounds.mes

DudeMcDude: Not sure about tpmes. I'm on vacation now, will get back to you next week.

PR #510, opened 2021-06-05
Secret Door port

anatoliy-savchak:
Port of Secret Door functions. No alterations.

### DIFF
--- a/TemplePlus/obj.cpp
+++ b/TemplePlus/obj.cpp
@@ -345,7 +345,6 @@ Objects::Objects()
 	
 	rebase(_IsPlayerControlled, 0x1002B390);
 
-	rebase(_SecretdoorDetect,	0x10046920);
 	rebase(_FadeTo,				0x1004C390);
 	rebase(_GetSize,			0x1004D690);
 
--- a/TemplePlus/obj.h
+++ b/TemplePlus/obj.h
@@ -213,10 +213,6 @@ struct Objects : temple::AddressTable {
 		_ContainerToggleOpen(handle);
 	}
 
-	bool SecretdoorDetect(objHndl door, objHndl viewer) {
-		return _SecretdoorDetect(door, viewer);
-	}
-
 	bool IsUndetectedSecretDoor(objHndl handle) {
 		auto flags = GetSecretDoorFlags(handle);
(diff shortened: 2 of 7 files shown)


anatoliy-savchak:
@doug1234, may I ask you to make code review? This is port from IDA, no changes.

doug1234:
One thing has me a bit confused.  What's up with SecretdoorDetect in obj.h?  Should it be changed to call you function directly?

anatoliy-savchak:
@doug1234, indeed there is no need now to use objects.SecretdoorDetect. Removed that and made reference to secret_door directly in python_object. Thank you!!

doug1234:
That change looks good.

DudeMcDude:
> @doug1234, may I ask you to make code review? This is port from IDA, no changes.

Hmm, did you make sure to include the fix for seen scenery?

anatoliy-savchak:
> Hmm, did you make sure to include the fix for seen scenery?
What fix?

DudeMcDude:
What I mentioned in co8. I want to make sure this doesn't undo the fix for secret doors. Are you sure it doesn't?

anatoliy-savchak:
@DudeMcDude, I have not touched that. Moreover there will be no changes in algorithms in comparison to previous implementation in toee.dll. I simply exposed them in T+.

This is your fix, right? It remains untouched.
```
uint32_t SecretDoorSys::secretDoorIsRevealed(objHndl secDoor)
{	// this function replacement should fix the motherfucking secretdoor bug
	// it worked on Aquinas' save!
	// The problem was that there's some sort of "revealed scenery" list which for some reason the secret door appears there
	// so I just removed that - there's the secret_door_found flag anyway
	// come to think of it, it may have supposed to prevent the player from searching again and again, but fuckit
	// they take 20 anyway (well maybe not in KotB >_>)
	if (!secDoor) return 0;
	uint32_t secDoorFlags = objects.getInt32(secDoor, obj_f_secretdoor_flags);
	if ( (secDoorFlags & OSDF_SECRET_DOOR) == 0 )  return 0;
	return (secDoorFlags & OSDF_SECRET_DOOR_FOUND) != 0;
	// there was also a look up in some name table, but that seems to be bugged (the door would already appear there in the secret door bug case, thus preventing its revelation)
}
```

PR #516, opened 2021-06-12
First part of the spell_compendium merge

Sagenlicht:
* Draft contains a new mod to spell_utils. I added the option to pass a parameter for tooltip and effect tolltip, so I can use the same  indicator file for the base spell and its mass version. This might be of interest to @dolio as well.

The spell Castigate works, but I am unsure if my code actually is unnecessary complex. Please review it.
 
Known Minor Issues:
* All DT_SONIC based spells should not pass through or work in a area of silence. I do not know if this is doable in ToEE atm.

Spells that use Vanilla conditions:
* Herald's Call (Slow)
* Invisibility, Swift (Invisibility)
* Ironthunder Horn (uses Knockdown)
* Ray of Light (Blindness)
* Frost Breath (Daze)
* Resist Energy, Mass (Resist Elements)

EDIT: Removed WF Ray Fix


DudeMcDude:
About the Weapon Focus Ray fix: that's nice, but it fails to address a related bug - if you pick WF Ray and WF Unarmed Strike, both will give you to hit bonuses.

IMO you can just completely replace the original Weapon_Focus and Greater_Weapon_Focus modifiers in python. They're very short, just 2 hooks so it should be simple. That way you also won't need to add this to every single ray spell in the game and worry about the expiring the condition.

Sagenlicht:
> About the Weapon Focus Ray fix: that's nice, but it fails to address a related bug - if you pick WF Ray and WF Unarmed Strike, both will give you to hit bonuses.
> 
> IMO you can just completely replace the original Weapon_Focus and Greater_Weapon_Focus modifiers in python. They're very short, just 2 hooks so it should be simple. That way you also won't need to add this to every single ray spell in the game and worry about the expiring the condition.

I fully agree with your statement, that WF Ray should be fixed instead of a workaround.

I just wasn't sure if this actually an option as spells like Scorching Ray have a fix for this (teds)

https://github.com/GrognardsFromHell/TemplePlus/blob/master/tpdatasrc/tpgamefiles/scr/Spell733%20-%20Scorching%20Ray.py

DudeMcDude:
Yes, obviously we'll have to undo all those manual interventions. But at least going forward all the new ray spells won't have to worry about that.

DudeMcDude:
So about castigate: personally I would have assigned numeric values as X-Y coordinates and checked the difference, but I assume what you did works. Other than that though, isn't it supposed to deafen enemies?

A couple of general remarks:
1. You should be sure you have enough args, because once you set the number of args you must not change it in the future, otherwise saves can get corrupted. For that reason it's ok to reserve a spare (or more), depending on the spell's complexity.

2. Wouldn't you like to do as dolio did and create a standard spell modifier that adds all the copy-paste hooks?

Sagenlicht:
Ok, numeric values. Noted. About deafen. No, but I forgot to check if the target can hear. The spell only works if target can actually hear the caster. Will drop the spell from the first merge.

1. Ok, one spare for normal, two for more complex? Will add them

2. I am not aware what dolio is doing in addition to spell_utils, will take a look and change the spells according to it then.

dolio:
I haven't taken a lot of advantage of it in what I've implemented so far, but `TouchModifier` that I added to spell_utils automatically does a lot of stuff that is relevant to most touch attack conditions. A similar thing could be done for other spells.

DudeMcDude:
Cool. So is this ready?

Sagenlicht:
> Cool. So is this ready?

Close to, I wanted to do the constants.py entries with the bonus_types [https://github.com/GrognardsFromHell/DllDocumentation/wiki/Bonus-Types](url) tomorrow and the merge them into the spells.
Also wanted to do a few particle effects for two or three spells. 
Fix missing sound effects.

Time frame: effects and sound until Wednesday, merge with bonus_types once bonus_types entries are approved. If you want to push the release of T+ I can stop working on deal_attack_damage and just finish particles and sound by tomorrow

DudeMcDude:
No rush. I also want to do a round of testing before pushing the version out since we changed a lot of things.

Sagenlicht:
> No rush. I also want to do a round of testing before pushing the version out since we changed a lot of things.

I just stumbled upon a major oversight by me. Spell stacking and duration refreshing. At the moment all my conditions are set to prevent duplicate, and while this is being correct in the sense that rules do not allow the stacking of the same spell twice it ignores the fact, that the second spell cast on the same target would ofc last longer and thus prolong the effect of the spell.

Now vanilla has not many spells that actually grant stacking bonuses and can simply set prevent duplicate to false due to this (e.g. Prayer, Bless, Doom).

I will simply do this with the spells from the spell compendium where this is applicable as well.

Sadly there are quite a few spells that actually grant stacking effects (e.g. Focusing Chant, which is a circumstance bonus). In addition alot of debuff spells have no type in the spell compendium and I used 0 as a bonus type due to this.

Now for those spells, I cannot simply set prevent duplicate to False. Is there a better way to handle this than using OnConditionAddPre?

tpdatasrc/tpgamefiles/scr/spell_utils.py
```diff
@@ -101,6 +101,15 @@ def removeTempHp(attachee, args, evt_obj):
     attachee.d20_send_signal(S_Spell_End, args.get_arg(0))
     return 0
 
+#Used to replace same condition to prevent duplicates
+def replaceCondition(attachee, args, evt_obj):
+    name = spellName(args.get_arg(0))
+    if evt_obj.is_modifier("sp-{}".format(name)):
```
DudeMcDude, line 107: You can get the condition name y'know

Sagenlicht:
@DudeMcDude 
I think I am finished and it is ready to merge.

tpdatasrc/tpgamefiles/scr/spell_utils.py
```diff
@@ -88,6 +101,15 @@ def removeTempHp(attachee, args, evt_obj):
     attachee.d20_send_signal(S_Spell_End, args.get_arg(0))
     return 0
 
+#Used to replace same condition to prevent duplicates
+def replaceCondition(attachee, args, evt_obj):
+    conditionName = args.get_cond_name()
+    if evt_obj.is_modifier("{}".format(conditionName)):
```
DudeMcDude, line 107: Is the format necessary here?

tpdatasrc/tpgamefiles/scr/tpModifiers/sp_appraising touch.py
```diff
@@ -0,0 +1,23 @@
+from templeplus.pymod import PythonModifier
+from toee import *
+import tpdp
+from utilities import *
+import spell_utils
+print "Registering sp-Appraising Touch"
+
+def appraisingTouchSpellBonusToAppraise(attachee, args, evt_obj):
```
DudeMcDude, line 8: Yknow, in ToEE this spell is pretty busted. Appraise has a wildly different meaning in ToEE than in P&P, direcrly translating to thousands of GPs or more.

Sagenlicht: I have no problem changing spells for balacing reasons.  Change it to  +1/casterlevel?

DudeMcDude: I think it's better to leave it out completely, since this starts getting too much into homebrew territory for my liking, which is better left for 3rd party mods.

Sagenlicht: Ok I will drop it then

DudeMcDude:
One more thing that I think is ok but I want to spell out the assumptions to be sure:
Spells with replaceDup call remove_spell
- For the simple case of single target spells that's straightforward and will end the spell
- For multiple target spells, this in general will not actually end the spell since the target list will not be fully cleared.
- This applies to the default spell remover. For emphasis- these new spells have no custom spell removal and rely on the default logic.

Sagenlicht:
This wouldn't be a problem, the whole idea behind this to allow to refresh spells before they actually expired. If only a few targets get refreshed that's fine.

What do think in general about this matter?

DudeMcDude:
Hmm, well there is something awkward about it.
On the one hand, you use untyped bonuses, which in principle allows stacking.
On the other hand, you disallow duplicates spell effects, which is a workaround for multiple spell castings to not stack.

So this begs the question, why not just assign a unique bonus type so that multiple castings just don't stack with each other (but do stack with other bonus types).
A related question - are multiple spell effects of the same type allowed? If so, are they supposed to confer some benefit for stuff like Dispel Magic? (e.g. you cast Prot vs. Evil twice on someone, enemy casts dispel -> you still have the other one active)

Sagenlicht:
I have thought about giving unique bonus types, so they  do not stack. I think that's the other option. I am not sure what option is the better one. Both require that you set the preventDup to False though.

I am no rules laywer, I can't answer your question about dispel magic with certainty. I always played it in my groups that the old spell is suppressed but yeah not 100% sure if it should count against dispel.

dolio:
I haven't done an exhaustive study, but the way e.g. bull's strength works is that it just allows multiple castings, since they don't stack. The question is, which things actually work that way in engine? For instance, if something adds extra damage dice, can you specify a type so that you only get the best dice?

On type of thing this definitely won't work for is event-based stuff. If you're allowed to cast disrupting weapon twice on an item, I don't think there's any way to get the kill-on-damage event to only trigger once.

DudeMcDude:
Alright, so it looks like unique bonus types is the preferable option - I think that’s how vanilla rolls for the most part as well.
As dolio mentioned, other effects such as extra dice etc aren't so easy to manage so use the dupe check method.

Sagenlicht:
Ok I will change it accordingly.

tpdatasrc/tpgamefiles/scr/spell_utils.py
Sagenlicht: Changed

Sagenlicht:
I added unique ID's to following spells:
ID | Spell
|----|:------------------------------|
155 | Sirenes Grace
156 | Sticky Fingers
157 | Undersong
162 | Curse of Ill Fortune
163 | Curse of Impending Blades
164 | Wave of Grief

I skipped both spells that have a circumstance modifier, in my opinion it looks nicer if the different sources are one sum. If I am only one, I change it.

Now that all curses have a unique modifier ID, I added checkCurseRemoval to spell_utils, the list should be centralized anyways.

@dolio 
I noticed, you modified the spell_utils in the some functions I also modified :(
I changed the tooltips, so I can use the tooltip icons (the indicator files) for the normal spells as well as the mass versions. If this is a bad idea, I would need to change the mass spells. 

Due to this, this cannot be merged right now.

tpdatasrc/tpgamefiles/scr/spell_utils.py
```diff
@@ -125,6 +150,36 @@ def skillCheck(attachee, skillEnum, skillCheckDc):
     checkResult = True if skillRollResult >= skillCheckDc else False
     return checkResult
 
+#Check if a spell that remove curses is cast
+#This list can be expanded if new curse removal spells are added
+#Limited Wish, Miracle and Wish also remove Curses but are not added to the game
+def checkCurseRemoval(evt_obj):
+    if (evt_obj.is_modifier("sp-Break Enchantment")
```
DudeMcDude, line 157: I absolutely want to avoid this kind of thing. I'd prefer modifiers that have this property register themselves.

Sagenlicht:
Removed checkCurseRemoval, Camouflage and Focusing Chant also have now new unique modifier types

dolio:
@Sagenlicht Sorry about that.

The way that you'd fix that on the PR side is to merge the master branch into your branch locally and then fix the conflicts during that merge.

I can help or do it if you want. (I'm not sure I have the permissions for the latter, though.)

Sagenlicht:
No need to be sorry dolio :) It's just we both wanna use the param for different options. Did you take a look on it? Do you think my idea is actually viable or should be discarded? I think we should discuss this first before we try to merge it.

dolio:
Oh, I see.

I don't have a strong feeling for how to fix it. If you want to make them separate functions, or make the functions recognize up to two parameters, either is fine.

DudeMcDude:
> @Sagenlicht Sorry about that.
> 
> The way that you'd fix that on the PR side is to merge the master branch into your branch locally and then fix the conflicts during that merge.
> 
> I can help or do it if you want. (I'm not sure I have the permissions for the latter, though.)

You've probably seen it by now, but I added you to the Modders group, which gives you write access to the repository. Use it well ;)

DudeMcDude:
@Sagenlicht congrats on the first merge :)
Could you add it to the changelog in the wiki?

Sagenlicht:
> @Sagenlicht congrats on the first merge :)
> Could you add it to the changelog in the wiki?

Ty :) and will do on Monday

PR #518, opened 2021-06-16
wf ray fix for feats

Sagenlicht:
added both feats as python feats
removed shiningteds workaround in rays


tpdatasrc/tpgamefiles/scr/tpModifiers/weapon_focus_ray.py
```diff
@@ -0,0 +1,25 @@
...
+    elif attachee.has_feat(feat_weapon_focus_unarmed_strike_small_being): #workaround to not stack with wf_unarmed
+        return 0
+    else:
+        bonusValue = 1
+        bonusType = 0 #ID 0 = Untyped (stacking)
+        bonusMesId = 114  #ID 114 in bonus mes: ~Feat~[TAG_FEATS_DES]
+        evt_obj.bonus_list.add_from_feat(bonusValue, bonusType, bonusMesId, feat_weapon_focus_ray)
+    return 0
+
+wfRayFix = PythonModifier("Weapon Focus Ray Fix", 0)
```
DudeMcDude, line 23: Feats should have 2 args. Also, users who have WF unarmed will still get the bonus on rays. I'll post the correction as an example for future reference.

PR #519, opened 2021-06-19
Update damage_critter.py

Sagenlicht:
@DudeMcDude 
EventObjDamage in tpdp misses a dispatcher, I think I would need this first, to get some testing going.


DudeMcDude:
Added. Also a bunch of other funcs you'll need. Hopefully I covered everything for damage_critter.

Sagenlicht:
Updated; Subdual damage TBD, I ran out of time.
It is working, damage is dealt twice and has same amount.
stubs:
*gameSystems->GetAnim().PushGoalHitByWeapon(attacker, tgt);
*tgt.sound_play_friendly_fire(?) - I don't know what arg to pass
*Set last hit

PR #520, opened 2021-06-19
Create generic_monster_dr.py

Sagenlicht:
Converted to Draft because my condition does not have the query all monster DR's have, e.g. https://github.com/GrognardsFromHell/DllDocumentation/wiki/Conditions-(List)-page-2-M-Z#monster-dr-holy

If I understand it correctly, It just checks if data1(which would be arg2 in my condition) has a valid (>0) value and if not it prints a debug msg

Should I add it to the generic_monster_dr.py or not? Can't see a real benefit.

### DIFF
--- a/tpdatasrc/tpgamefiles/scr/tpModifiers/generic_monster_dr.py
+++ b/tpdatasrc/tpgamefiles/scr/tpModifiers/generic_monster_dr.py
@@ -1,0 +1,22 @@
+from templeplus.pymod import PythonModifier
+from toee import *
+import tpdp
+
+### Generic Monster DR Condition
+
+def grantMonsterDr(attachee, args, evt_obj):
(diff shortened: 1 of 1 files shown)


DudeMcDude:
Not sure what exactly it's for but let's be conservative with it and keep it.

Sagenlicht:
> Not sure what exactly it's for but let's be conservative with it and keep it.
> 
> I do see these errors sometimes btw.

I just tested target.d20_query(S_Verify_Obj_Conditions) in game with a plain Vrock.  Wanted to see the return before I start to implement the hook into the generic_monster_dr. 

The game crashes :(

DudeMcDude:
Oh, I hadn't even noticed it was query rather than signal. Heh. It should be a signal I suppose.
But I guess since it's buggy in itself, just leave it out.

PR #525, opened 2021-06-27
Replicate some engine AI functionality in Python

dolio:
I'm starting in on the `ConsiderTarget` replication, but have some questions, and thought just using a preemptive draft to ask them is as good as anything.

1) Where should I put the file? I thought maybe `.../d20_combat/ai.py` but wasn't certain.
2) One thing I need to replicate is comparing an object to the result of `d20_query_get_data(Q_Critter_Is_Charmed, 0)`. Is there a way to turn a numeric handle into its object or vice versa? I thought I'd seen this, but maybe I'm just thinking of `obj_get_obj`, and I need the `d20_query` analogue.

I'm sure I'll have additional questions later.


DudeMcDude:
1. Put it in d20_ai (since in principle I is not just about combat).
2. In python, the result of d20_query_get_data and obj.get_obj return object handles, so you can just use the usual comparison operators.

I've also started a branch for this - PythonAI.

dolio:
No, I'm not using `obj_get_obj`. The issue is that presumably for `isCharmedBy(x,y)`, `x` and `y` are actual python objects, but the d20_query only returns the handle as a long.

By the way, is there some kind of function naming convention that you prefer? It seems like it's just all over the place in a lot of these files (underscores, camel case, capitalized, ...).

dolio:
Further question: any thoughts on how far I should drill into the logic?

I can keep replicating some of the stubbed out stuff in Python, but it might not be worth it. Some stuff, like `AiCannotHear` involves raycasting that might be more work to expose than just the whole function.

tpdatasrc/tpgamefiles/rules/d20_ai/targeting.py
```diff
@@ -24,25 +24,82 @@ def considerTarget(attacker, target):
...
+
+	return kos_candidate
+
+# Common check during findSuitableTarget
+def detect(attacker, target):
+	if target == OBJ_HANDLE_NULL: return False
+
+	unconcealed = isUnconcealed(target)
+
+	canHear = missing_stub("CanHear(attacker, target, unconcealed)")
```
DudeMcDude, line 102: These are implemented in the python api as has_los and can_hear

DudeMcDude:
> No, I'm not using `obj_get_obj`. The issue is that presumably for `isCharmedBy(x,y)`, `x` and `y` are actual python objects, but the d20_query only returns the handle as a long.
> 
> By the way, is there some kind of function naming convention that you prefer? It seems like it's just all over the place in a lot of these files (underscores, camel case, capitalized, ...).

Oh sorry, I didn't see your reply.
So yeah looks like we'll have to add d20_query_get_obj. (The alternative is to change _get_data to return longlong, and let you instantiate PyObjHandle from it, but I don’t want to risk breaking other things)

For naming, prefer to use lower camel for func names.

tpdatasrc/tpgamefiles/rules/d20_ai/targeting.py
DudeMcDude: Well just has_los actually. I think the return vakue is inverted though. Will add can_hear.

dolio:
I think `d20_query_get_data` already returns the packed longlong, actually. At least, it seems to me that way. I'd thought that the second argument might select from `data1` or `data2`, but it seems to return them both regardless.

But `d20_query_get_obj` would be a preferable interface anyway. I guess some way to get a python object from a handle number or vice versa might be handy, though. I've noticed some other API holes that I can't work around without that in fiddling with other stuff. For instance, there are methods `get_obj_from_args` and `set_args_from_obj`, but there's no way to `condition_add` that allows object arguments (I think).

DudeMcDude:
> I think `d20_query_get_data` already returns the packed longlong, actually. At least, it seems to me that way. I'd thought that the second argument might select from `data1` or `data2`, but it seems to return them both regardless.

Oh well in that case I think you can instantiate with PyObjHndl(numberHandle).

> But `d20_query_get_obj` would be a preferable interface anyway. I guess some way to get a python object from a handle number or vice versa might be handy, though. I've noticed some other API holes that I can't work around without that in fiddling with other stuff. For instance, there are methods `get_obj_from_args` and `set_args_from_obj`, but there's no way to `condition_add` that allows object arguments (I think).

Well you could convert the handle to two ints. A bit hacky though, I suppose it's better for the python API to handle that internally.

dolio:
> Well you could convert the handle to two ints. A bit hacky though, I suppose it's better for the python API to handle that internally. I'll add that as well.

I can? Is it like `long(obj)` and then shift/mask or something? I also didn't know about `PyObjHndl(num)`. I was trying to find some indication of how to do that, but the C++ side implementing that API doesn't really have any clear indication that it's possible.

That's handy if it works, because sometimes I'm just messing around with an idea, and it'd be nice if I could manually do the hacky solution to get something working, even if it's subsequently be replaced by a proper API call.

DudeMcDude:
> > Well you could convert the handle to two ints. A bit hacky though, I suppose it's better for the python API to handle that internally. I'll add that as well.
> 
> I can? Is it like `long(obj)` and then shift/mask or something? I also didn't know about `PyObjHndl(num)`. I was trying to find some indication of how to do that, but the C++ side implementing that API doesn't really have any clear indication that it's possible.
> 
> That's handy if it works, because sometimes I'm just messing around with an idea, and it'd be nice if I could manually do the hacky solution to get something working, even if it's subsequently be replaced by a proper API call.

You can check out how Co8 did it in co8util/ObjHandling.py.

dolio:
Oh, they parse it out of the to-string output For some reason I didn't think of that. :smiley:

PR #526, opened 2021-07-07
Update damage_critter.py

Sagenlicht:
Added deal_spell_damage to damage_critter.py
Fixed some errors in deal_attack_damage function
moved set_weapon_used and float friendly fire to own functions as it used by deal_attack_damage and deal_spell_damage
@DudeMcDude 
I would need a hook to test deal_spell_damage :)


tpdatasrc/tpgamefiles/rules/d20_combat/damage_critter.py
```diff
@@ -32,28 +51,16 @@ def deal_attack_damage(attacker, tgt, d20_data, flags, action_type):
     
     #Create evt_obj_dam
     evt_obj_dam = tpdp.EventObjDamage()
-    evt_obj_dam.attack_packet.action_type = action_type
+    missing_stub("evt_obj_dam.attack_packet.action_type = action_type #this returns an error")
```
DudeMcDude, line 54: Eh? Didn't this function work before?

tpdatasrc/tpgamefiles/rules/d20_combat/damage_critter.py
```diff
@@ -128,7 +134,7 @@ def deal_attack_damage(attacker, tgt, d20_data, flags, action_type):
     damage_critter(attacker, tgt, evt_obj_dam)
 
     #Play damage particle effects
-    evt_obj_dam.damage.play_pfx(tgt)
+    evt_obj_dam.damage_packet.play_pfx(tgt)
```
DudeMcDude, line 137: ??? Same here

tpdatasrc/tpgamefiles/rules/d20_combat/damage_critter.py
Sagenlicht: It never worked on my side, I thought you might have fixed it, when I saw it unstubbed. But I still get the same error when testing it.

tpdatasrc/tpgamefiles/rules/d20_combat/damage_critter.py
Sagenlicht: I have fixed this, you added the function play_pfx in tpdp and unstubbed it.

The problem is damage (which is correct in c++)  but needs to be damage_packet in python.

Sagenlicht: To be clear, I tested this yesterday quite a bit, atm the deal_attack_damage function basically works as it should expect the stubbed parts (action type and a sound call). The damage looks fine.

Testing environment: I add the condition Unconscious to a party member, let him fall_down()  and then a second party member uses Coup de Grace.

What I haven't tested so far is the concealment and the deflect arrow part, which is on my ToDo for today.

DudeMcDude: Hmm ok, I recall testing it but maybe I added these afterwards. Will check.

Sagenlicht: Ty :)
If you find time, could you please also add a hook to the deal_spell_damage so I can start testing it? I think it looks pretty good, only stub I have (besides the action_type) would be the Metag Magic

DudeMcDude: I have a proposal for you ;) can you copy the stuff I did in deal_attack_damage?

Sagenlicht: Just to be sure, I need both correct?

https://github.com/GrognardsFromHell/TemplePlus/blob/master/TemplePlus/damage.cpp#L421
https://github.com/GrognardsFromHell/TemplePlus/blob/master/TemplePlus/damage.cpp#L537-L549

DudeMcDude: Yeah

Sagenlicht:
Currently know Issues:

Friendly Fire Float is not working
PyObjHandle_AllegianceShared (attacker.allegiance_shared(tgt) returns 0
Tested with female dwarf and human male party members and with two male human party members

deal_attack_damage:
Coup de Grace completly ignores concealment, there is no concealment roll at all
Tested with Displacement on a unconscious party member

Missing stubs:
Meta Magic in deal_spell_damage

Untested:
Deflect Arrows in deal_attack_damage

Miscellaneous:
print of evt_obj_dam.dispatch_spell_damage(attacker, tgt, spellPacket) retuns 0 but spell damage is dealt normally

DudeMcDude:
> The C++ re-checks for primary/secondary weapon, but it doesn't check for touch attacks (as far as I can tell). I would guess this is because touch attacks don't deal damage themselves, they're usually part of some kind of ability.
> 
> I think the answer to my earlier question is that the damage function doesn't have the original attack packet (I thought it did), so it can't get the weapon from there.
> 
> However, rather than just suggesting to roll it back, I guess I was asking a meta-question about what `D20CAF_TOUCH_ATTACK` should indicate. Is it indicating a type of attack action (and thus weapon used)? Or is it about the to-hit properties of any attack. Right now it kind of does both, which is a problem if you try to use it in both ways
> 
> I'm kind of leaning toward it being the attack _action_ flag, and weapons that strike as touch attacks are better implemented as AC bonus caps like you originally did (sorry).
> 
> BTW, I think this change doesn't affect anything in the existing game. It only came up for me because I was fiddling around with Flame Blade.

Ah, I remembered we had discussed this. Turns out that this addition is a bug - it causes thrown grenades attacks (flagged as D20CAF_TOUCH_ATTACK) to deal unarmed damage instead of whatever they were supposed to.

That also settles the usage intent - it's for the to-hit properties of the attack.

PR #531, opened 2021-07-10
Disrupting Weapon

dolio:
Here's another thing I've been working on to learn how item enhancements work. It implements the [Disrupting Weapon](https://www.d20srd.org/srd/spells/disruptingWeapon.htm) spell.

There's 2 problems, though.
1. I can add the relevant effect to the weapon with `item_condition_add_with_args`, but it seems there's no way to undo this when the spell ends. So the effect is permanent. That isn't a problem for effects for crafted/proto weapons (and it's how I got the idea that Fragarach could be re-implemented), but it doesn't work for limited time spells.
2. Applying the above condition doesn't refresh the character's status if they're already holding the weapon. The C calls a function to reset all item-based status, but that doesn't seem possible in python. I tried `d20_status_init`, but that seems like it's not the full refresh, and probably only the item stuff should be reset anyway.

It's a core spell, so I put it in the first apparently unused core slot I could see (590), but I can move it if that's a faux pas.


DudeMcDude:
1. Yes, this is something that needs to be added. Internally what the engine does is just delete some data in obj_f_item_pad_wielder_condition_array and obj_f_item_pad_wielder_argument_array (see SpellRemoveItemCond; actually it's buggy since it just removes the end of the array)
2. Yes, d20_status_init only works for objects that haven't been inited yet (have a null dispatcher). I can add d20_status_init_items(), and in the meantime you can unequip (.item_worn_unwield) and re-equip it so the internal functions handle it.

@Sagenlicht this is relevant for you too.

Sagenlicht:
> 1. Yes, this is something that needs to be added. Internally what the engine does is just delete some data in obj_f_item_pad_wielder_condition_array and obj_f_item_pad_wielder_argument_array (see SpellRemoveItemCond; actually it's buggy since it just removes the end of the array)
> 2. Yes, d20_status_init only works for objects that haven't been inited yet (have a null dispatcher). I can add d20_status_init_items(), and in the meantime you can unequip (.item_worn_unwield) and re-equip it so the internal functions handle it.
> 
> @Sagenlicht this is relevant for you too.

Oh this would be great news, I actually use this workaround atm!

def sonicWeaponSpellAddWeaponCondition(attachee, args, evt_obj):
    mainhandWeapon = attachee.item_worn_at(item_wear_weapon_primary)
    mainhandWeapon.item_condition_add_with_args('Weapon Sonic', args.get_arg(1))
    ####   Workaround to activate weapon enchantment   ####
    attachee.item_worn_unwield(item_wear_weapon_primary)
    attachee.item_wield(mainhandWeapon, item_wear_weapon_primary)
    #### Workaround to activate weapon enchantment end ####

DudeMcDude:
> > 1. Yes, this is something that needs to be added. Internally what the engine does is just delete some data in obj_f_item_pad_wielder_condition_array and obj_f_item_pad_wielder_argument_array (see SpellRemoveItemCond; actually it's buggy since it just removes the end of the array)
> > 2. Yes, d20_status_init only works for objects that haven't been inited yet (have a null dispatcher). I can add d20_status_init_items(), and in the meantime you can unequip (.item_worn_unwield) and re-equip it so the internal functions handle it.
> > 
> > @Sagenlicht this is relevant for you too.
> 
> Oh this would be great news, I actually use this workaround atm!
> 
> https://github.com/Sagenlicht/ToEE_Mods/blob/main/overrides/scr/tpModifiers/sp_sonic_weapon.py#L8-L14

Ok I've changed it so item_condition_add will also try to refresh the item conditions for the item's parent. Seems to be the only relevant use case.

dolio:
That seems fine, as long as whatever removes it does the same (I forgot to mention that part).

Sagenlicht:
Hey @dolio,
if this is of interest to you this is how my weapon enchantments look like after the new update:
[https://github.com/Sagenlicht/ToEE_Mods/blob/main/overrides/scr/tpModifiers/sp_align_weapon.py](url)

dolio:
@Sagenlicht So, maybe I'm reading it wrong, but this actually doesn't work the way I wanted mine to work.

I mean, the difference is that it will actually work correctly currently as a timed effect.

However, it doesn't allow you to align a weapon and then have that weapon moved to a different critter and still work appropriately, because the condition that makes it work is tied to the original target of the spell. That condition will stay on the original target regardless of who holds the weapon. That's the purpose of `item_condition_add_with_args`, to have an effect applied on a critter while the item is equipped (or, I think as long as it's held for certain inventory spell objects).

I think your approach probably covers most situations that matter (like, disarming someone and then picking up and using their weapon is not going to happen very much), since your spell doesn't require going through the 'cast in inventory then transfer' thing. But I wanted to replicate the way e.g. Magic Weapon works.

Sagenlicht:
> @Sagenlicht So, maybe I'm reading it wrong, but this actually doesn't work the way I wanted mine to work.
> 
> I mean, the difference is that it will actually work correctly currently as a timed effect.
> 
> However, it doesn't allow you to align a weapon and then have that weapon moved to a different critter and still work appropriately, because the condition that makes it work is tied to the original target of the spell. That condition will stay on the original target regardless of who holds the weapon. That's the purpose of `item_condition_add_with_args`, to have an effect applied on a critter while the item is equipped (or, I think as long as it's held for certain inventory spell objects).
> 
> I think your approach probably covers most situations that matter (like, disarming someone and then picking up and using their weapon is not going to happen very much), since your spell doesn't require going through the 'cast in inventory then transfer' thing. But I wanted to replicate the way e.g. Magic Weapon works.

I didn't test it with Align Weapon, but Sonic weapon actually adds the damage when it is wielded by another character.
https://github.com/Sagenlicht/ToEE_Mods/blob/main/overrides/scr/tpModifiers/sp_sonic_weapon.py
What does not work indeed is the Buff symbol when moved to another character as it is tied to the original spell target. 
In the beginning I used to do the item enchantments the way magic weapon works (target mode: item) but I redid them all so they automatically enchant the primary weapon of a normal target. Yes, this is a limitation, but this makes mass spells so much easier to handle (like mass align weapon). You can switch the weapon to the offhand afterwards if really needed.

I am not 100% sure that my approach is actually the best way to do it, but atm it's the best way to get item enchantments to work (I believe), though if it's not I would be glad if I you could point me how to do it better.

I think I have an idea how I can fix the tooltip matter as well (I'd like to have it).

I agree, an item_condition_remove would be perfect and solve some problems with item enchantments.

dolio:
@Sagenlicht Your sonic weapon works because you're using `item_condition_add`. for the 'Weapon Sonic' condition. That is the one that adds a condition that applies to the character while they're holding the weapon. So, while they're holding the weapon, the damage and glow checks will happen, and it will tick.

However, nothing will ever remove that item condition from the item. It might expire from the _character_ once it ticks down the first time, but if you re-equp it, I think the effect will be active again. The counter might be the same as last time, so it will expire after 1 round, but I would guess that fiddling with equipment will reactivate it for one round.

Another thing to note: when using `item_condition_add`, you don't need to program your own checks for the appropriate weapon. This is because (as long as there are 3+ arguments) the engine automatically sets arg 2 of the condition to the 'inventory number' of the item it came from. So when checking for glow or damage, you can check if the weapon in `evt_obj` has the same inventory number as `args.get_arg(2)`. This made things a lot simpler than when I thought I had to implement my own python queries. You can see the code in this pull request for how to use it.

DudeMcDude:
The Long Description now fetches item enhancements automagically. See LongDescriptionPopupCreate in ui_char.cpp and UiItemCreation::GetEffectDescription.

DudeMcDude:
About Sonic Weapon: the convention for item enchantments seems to be 5 args, with the following specs:

args[0] = bonus value (for variable bonus conditions Weapon Enhancement Bonus)
args[1] - not sure what in general, but for Defending weapon it's used for the set bonus amount
args[2] = inventory idx (in general for item conditions)
args[4] = spell id

Dunno abotu [3]

dolio:
Some existing item effects only have 3 arguments, presumably because they're never expected to be added as part of a spell. There is some logic related to whether or not the condition has 5 arguments, but I didn't bother to figure out exactly what it was doing.

I guess a question is: will Temple+ work for item conditions with more than 5 arguments (in case something is that complicated)? In the original code there's a function with 5 integer arguments that handles this stuff, so presumably it wouldn't there.

DudeMcDude:
More than 5 args shouldn't be a problem.
I do want to require at least 5 args as above, since I want to match the spell id when removing the condition.

Sagenlicht:
Ok I will fix all my item enchantment spells. Can I use arg[3] or should I set it 0?

dolio:
The only argument set in the function that sets arg 2 is 2. And 'Boots of Speed' seems to use arg 3 for its own purposes. So it seems like an available slot.

DudeMcDude:
Added item_condition_remove. As discussed, it also refreshes the item conditions on the item's parent.

DudeMcDude:
BTW I've also merged some preliminary python GUI stuff. You can do basic stuff like create windows with images / text, show/hide etc. Take a look and start thinking about use cases :)

Sagenlicht:
@dolio 
Wouldn't be verifyItem a great addition for spell_utils? I actually have quite a few item enchantments in the spell compendium.

@DudeMcDude 
Great news :)

anatoliy-savchak:
I would love to create Craft Masterwork Items dialog.

DudeMcDude:
Contingency/spell trigger etc too 😉

dolio:
Yeah, I've been wanting to try those, as well.

Also Mordenkainen's Lucubration which I find interesting because it's Wizard only, not Sorcerer.

That reminds me of another thing. I was looking around a while back with regard to Rary's Mnemonic Enhancer (another Wizard only). One of its effects basically gives temporary bonus spell slots. Is it possible to do that? It seemed like maybe they had plans for that at some point with one of the unused events (maybe 57 or 58; I don't exactly recall) but it got dropped. Bonus slots would also enable e.g. Precocious Apprentice.

doug1234:
Contingency for sure.  Could also allow selecting extra favored enemies with Swift Hunter.

DudeMcDude:
Just to verify, you're aware the game has 'Weapon Disruption', right?
Though I guess it doesn't matter much, since only St. Cuthbert's mace has the property, and it hasn't been craftable so far so there won't be conflicts.

dolio:
I know, I studied it to find some of the mes stuff and particle effects. I can't use it, though, because it's always DC 14, and it's 3 arguments, so can't be used for a spell. I could check for the condition in the spell before applying, but I guess it isn't really important?

I think it should be ready now, other than that.

Sagenlicht:
I would have two small questions

1. The spell expiry float is ugly, as it is not centered above the wielder, it floats somewhere else in the room.
2. My weapon enchantments do not show up in the weapon tooltip (the shift+click one). How do you want to handle this @dolio? Skip it?

DudeMcDude:
Temporary enhancements don't have to show up in the long description popup. Permanent ones have entries in the crafting rules IIRC.

dolio:
I asked about the shift-click thing above. It'd be nice for it to show up in there, but I don't think there's a way to do it right now.

The floated message also doesn't seem under my control. Magic Weapon does the same thing, for instance. I would guess it has to do with the inventory targeting mode or something.

DudeMcDude:
Oh damn, looks like Rudy's Mod uses spell 590 for Resist Acid.

Also there were quite a few tangent discussions here and elsewhere, they get their own threads / issues if you don't want me to forget about them ;)

dolio:
Okay, I can move it. I tried looking for that mod, but didn't see anything that would have conflicted, so I guess I found the wrong version or something.

Are there any free numbers that count as core spells? There are others that are missing from Troika's list. Maybe a new range can be designated? I think a couple previously implemented core spells might already be disabled if you use the core-only option, because they're 1000+. I didn't check, though.

DudeMcDude:
Hmm, might be the Domain Spells mod.

I don't mind lumping a few core spells with the non-core option. If it gets out of hand that can be reworked.

PR #533, opened 2021-07-11
Fix 'Wall of Fire' duration

dolio:
- By the book, it's 'Concentration + 1 round/level', rather than 'Concentration up to 1 round/level'. Basically you can cast it as a 'fire and forget' spell, but maintaining concentration extends it further.
- Includes some spell_utils additions to facilitate these sorts of  spells.

### DIFF
--- a/tpdatasrc/tpgamefiles/scr/spell_utils.py
+++ b/tpdatasrc/tpgamefiles/scr/spell_utils.py
@@ -30,17 +30,48 @@ def spellTime(duration):
 
 ### Standard Hooks invoked with AddHook ###
 
-#[pytonModifier].AddHook(ET_OnGetTooltip, EK_NONE, spell_utils.spellTooltip, ())
+# Handles tooltip requests for a condition. Can be installed as:
+#
+#   [pytonModifier].AddHook(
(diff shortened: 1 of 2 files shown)


DudeMcDude:
Thanks. Are you looking into wall spells?

dolio:
Actually it was the 'concentration + X' that I wanted to hash out, since I've seen other spells, and then noticed it applied to this one.

I thought about walls a little, but it seems like most of them would require spawning something to actually block movement, and that seems more complicated than I want to dig in to right now. Like, I assume it would be complicated by either having to know of the right proto that already exists, or creating an appropriate one with all the 'what about other modules' complications that entails.

Actually, even this spell does not technically match the SRD (even ignoring the alternate shapes). It seems like the wall part is almost supposed to be a 2D plane, and then one side of the wall blasts out to a distance of 20ft (with lower damage 10-20 ft, higher 0-10 ft, and the highest for actually going through the 2D part). But I think implementing that would significantly complicate the targeting options, and probably isn't worth it for this one spell.

I guess the 'radius' could also be increased. Making it 20ft makes it hard to even use in a lot of places, though. It gets hung up on level geometry really easily. You can basically fill an entire temple corridor with fire, and it'd be ridiculous on the fire node with double size.

PR #537, opened 2021-07-15
Remove 3.0 limitation of medium size for 'person' targets

dolio:
Several 3.0 spells had limitations on size rather than just humanoid targets. Since ToEE started as a 3.0 game (I think) these got written and never reviewed in the rules update.

This is a problem in that Enlarge Person increases your size (which it didn't in 3.0), and makes you immune to these spells, including Reduce Person, which is supposed to dispel it.

I also fixed some problems with Enlarge/Reduce. There were code paths that would leave the spell dangling forever (when one dispels the other). Enlarge also had some missing logic apparently from the Co8 rewrite to fix the size/reach stuff for PCs.

There are actually other spells with this logic.
1. Greater Command: The by-the-book limitation on the spell is that the caster needs to know a language that the target understands, but I don't think that's something that ToEE really tracks. The lesser version allows any humanoid or <large creature to be a target, but Greater just allows <large.
2. Mass Hold Person: This one seemed a little less urgent since it's high level. I also noticed that the targeting is wrong. It's doing 30ft radius enemies instead of Any 30 Feet (which is more like a 15ft radius). I can fix one or both of these, but I'm not sure which folders they should go in. Both co8infra and kotbfixes?

One thing I should note is that the fixed endless spell case for Enlarge Person causes complaints in the console because of its reach fix. However, I was wondering if maybe the reach calculation should just be replaced in Temple+ so that that fixup isn't necessary anymore. Reach is also not correctly calculated when you reduce a small character (they should have 0 natural reach, but it's set to 5).

### DIFF
--- a/tpdatasrc/co8infra/rules/spells/588 - Mass Hold Person.txt
+++ b/tpdatasrc/co8infra/rules/spells/588 - Mass Hold Person.txt
@@ -1,0 +1,21 @@
+School: Enchantment
+Subschool: Compulsion
+Descriptor: Mind-Affecting
+Level: Sor 7
+Level: Wiz 7
+Component: V
+Component: S
+Casting Time: 1 action
+Range: Medium
+Saving Throw: Willpower
+Spell Resistance: Yes
+Projectile: No
+flags_Target: Range
+inc_flags_Target: Self
+inc_flags_Target: Other
+exc_flags_Target: Dead
+exc_flags_Target: Non-critter
+exc_flags_Target: Friendly
+mode_Target: Area
+radius_Target: 30
+ai_type: ai_action_offensive
--- a/tpdatasrc/co8infra/rules/spells/588 - Mass Hold Person.txt
+++ b/tpdatasrc/co8infra/rules/spells/588 - Mass Hold Person.txt
@@ -15,7 +15,11 @@ inc_flags_Target: Self
 inc_flags_Target: Other
 exc_flags_Target: Dead
 exc_flags_Target: Non-critter
-exc_flags_Target: Friendly
-mode_Target: Area
-radius_Target: 30
+mode_Target: Multi
+mode_Target: Once-Multi
+mode_Target: Any 30 Feet
+mode_Target: End Early Multi
+radius_Target: 0
+min_Target: 1
+max_Target: 99
 ai_type: ai_action_offensive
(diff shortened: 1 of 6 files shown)


DudeMcDude:
> One thing I should note is that the fixed endless spell case for Enlarge Person causes complaints in the console because of its reach fix. However, I was wondering if maybe the reach calculation should just be replaced in Temple+ so that that fixup isn't necessary anymore. 

No objections here if you want to do it. One thing to note is that a redone fix should consider save games that have the previous method going.

dolio:
Okay, I'll think about how to accomplish that.

dolio:
Okay, how about this:

Simultaneously with whenever the automatic reach behavior is implemented, the enlargement spells are updated so that the OnBeginRound function undoes the obj_set_int version. Then there is at most one round of being too large. The code for removing things from the Co8 enlargement list can stay indefinitely, it just shouldn't print error messages when a spell ends with something not in the list.

DudeMcDude:
> Okay, how about this:
> 
> Simultaneously with whenever the automatic reach behavior is implemented, the enlargement spells are updated so that the OnBeginRound function undoes the obj_set_int version. Then there is at most one round of being too large. The code for removing things from the Co8 enlargement list can stay indefinitely, it just shouldn't print error messages when a spell ends with something not in the list.

Ok, but please also include some a safeguard against reducing reach twice, since the co8 permanent data is kinda wonky and beginround events can fire more than once.

Btw re. 99 targets. The spellpacket struct has a 32 targets capacity.

dolio:
Is it at least reliable that one `OnBeginRound` will happen completely before the next? So that if one reduces reach and removes the character from the permanent data, the next one will see that the permanent data does not contain them?

DudeMcDude:
I think so.

dolio:
Actually, looking more closely at the co8 thing, it can't double reduce/increase reach. What it actually does is calculate modified reach based on `obj_f_size`, which doesn't change with enlarge/reduce. Then the 'undo' just sets the normal reach for that size. It never uses the current reach to calculate the new reach.

There's some commented code for modifying the size category, but they figured out that's wrong anyway, because there are hooks that modify the stat.

DudeMcDude:
Is this PR complete, or did you want to add the Hold Person do-over?

dolio:
I think it's complete.

Not sure what you're referring to. I did all the hold persons, I think. If you mean the enlarge/reduce reach stuff, I'll do that separately, since it probably requires some C++ tweaking.

DudeMcDude:
> I think it's complete.
> 
> Not sure what you're referring to. I did all the hold persons, I think. If you mean the enlarge/reduce reach stuff, I'll do that separately, since it probably requires some C++ tweaking.

Yeah I meant enlarge/reduce person.

PR #550, opened 2021-07-25
Divine spell power was doubling caster level

dolio:
- The written feat adds the turn undead modifier to the caster level,
  not the level of turnable monsters.
- Seems like probably a left-over += from before the check to avoid
  non-positive caster levels was added.


doug1234:
It seems like this change would break all other sources of caster level bonus if divine spell power is active.

dolio:
They do in most cases.

I think there are corner cases where it might not work. Like Cleric 1/Other N with Practiced Spellcaster. Then whether or not the penalty gets applied depends on whether Divine Spell Power or Practiced Spellcaster gets called first. To avoid that you'd need something more like a `BonusList` with a minimum clamp. But I'm not sure it's worth worrying about that, unless someone is digging into that anyway (doesn't seem like an easy fix).

doug1234:
OK.  Sounds reasonable.

PR #554, opened 2021-07-29
WIP: Implosion

dolio:
I've been mostly avoiding high level spells since they don't seem like they'd get a ton of play, but I figured I'd try this one out for a few reasons:

1. I wanted to learn to do d20 actions, and this one lets you do an action each round
2. Concentration is actually kind of wrong in ToEE. It's supposed to require a standard action each round, but the way it's implemented is that it's free, and some actions that cost would cost too much break it. So this is kind of an 'active concentration' trial. I think the normal concentration could be fixed, but this spell has an associated action anyway.
3. Balors

One thing I'm missing is the ability to test if a D20Action is marked with the flag D20ADF_Breaks_Concentration. I think I need to do it in a hook. Spell casting automatically sends a signal to break concentration, but I think not every action does. Some of them might only be marked that way because they're standard actions, but it's possible to have a swift action that breaks concentration even though there is still enough time to concentrate in the round (spell casting is such an example, but I think it works fine).

Also, do I have all the files in the right spot? I was second guessing myself looking at prior stuff, and I still don't understand all the logic of which things go where. Some stuff seems to be in co8 but not kotb, but then other stuff is in both, and I'm wondering if I messed that up on e.g. Disrupting Weapon.


dolio:
Oh, also, I've been toying with the idea that this destroys the target, similar to disintegrate, but was trying to avoid using the Animate Dead condition. Is there a way to replicate the inventory dropping that that does in python? I know you can make people drop worn items, but I don't see how to force them to drop from the rest of the inventory.

DudeMcDude:
> Oh, also, I've been toying with the idea that this destroys the target, similar to disintegrate, but was trying to avoid using the Animate Dead condition. Is there a way to replicate the inventory dropping that that does in python? I know you can make people drop worn items, but I don't see how to force them to drop from the rest of the inventory.

Currently, no.

DudeMcDude:
> I've been mostly avoiding high level spells since they don't seem like they'd get a ton of play, but I figured I'd try this one out for a few reasons:
> 
> 1. I wanted to learn to do d20 actions, and this one lets you do an action each round
> 2. Concentration is actually kind of wrong in ToEE. It's supposed to require a standard action each round, but the way it's implemented is that it's free, and some actions that cost would cost too much break it. So this is kind of an 'active concentration' trial. I think the normal concentration could be fixed, but this spell has an associated action anyway.
> 3. Balors
> 
> One thing I'm missing is the ability to test if a D20Action is marked with the flag D20ADF_Breaks_Concentration. I think I need to do it in a hook. Spell casting automatically sends a signal to break concentration, but I think not every action does. Some of them might only be marked that way because they're standard actions, but it's possible to have a swift action that breaks concentration even though there is still enough time to concentrate in the round (spell casting is such an example, but I think it works fine).
> 
> Also, do I have all the files in the right spot? I was second guessing myself looking at prior stuff, and I still don't understand all the logic of which things go where. Some stuff seems to be in co8 but not kotb, but then other stuff is in both, and I'm wondering if I messed that up on e.g. Disrupting Weapon.

Concentration - yeah that needs to be added to API.
Every action sequence sends S_Sequence (see curSeqNext()), which is what the sp-Concentraion reacts to.
KotB support is a but of a mess. I had intended to separate ToEE specific and Co8 infrastructure, but then KotB has a bunch of module specific stuff in e.g. spell scripts, mes files etc. In this particular case, no harm done in excluding a 9th level spell from a levels 1-5 module ;)

DudeMcDude:
About action enum: we should coordinate which range you can use. Generally I reserve class_enum * 100 for each class, so 900 could be used by Clerics in principle. I'l get back to you on that.

dolio:
Yeah, I figured since it's a cleric spell, it might make sense to put it there.

I looked at the existing actions, and some seem to be slotted into classes that don't really have other abilities. Like divine armor/vigor/... are 260x, which is Eldritch Knight. I have some other spells on my list that I think would require some custom actions, and I don't know if it makes sense to put those under Wizard/Sorcerer, or if you want to reserve some separate block or something.

DudeMcDude:
Hmm, about concentration - what's the reason for creating a custom handler? Usually spells add the sp-Concentrating condition.

dolio:
The reasoning goes something like this:

1) Implosion already needs its own action for the per-round effect.
a) You could maybe try to require selecting all 4 targets up front, and do one each round, but that means you can't adapt to changes in the situation.
2) It's more important that the standard action cost of concentration is enforced, because it's a powerful effect. This is easily done by making the above cost a standard action, and count as your concentration.
3) I didn't know you could extend existing conditions in Python at the time, so I didn't know the timeline for making concentrating on a spell cost a standard action. And maybe that should be only a strict rule in general, but implosion shouldn't be free even in non-strict rules (_I_ think).
4) If you _try_ to fix it by making sp-Concentrating deduct a standard action, you run into corner cases:
a) If concentration removes your standard action, you can't do the implosion action each round unless it's free. The action system has no way of making the action free only for the round the spell is cast. Making Implosion a swift action spell to get around this also isn't correct.
b) If the action is free, you could do two implosions on the first round if it worked the current way, so you'd need to make it like a personal spell and make someone do the free implosion action the first round. That's not a problem, but it's more menuing.
c) If breaking concentration refunds the standard action, then you could get a free implosion on the last round by imploding first and then breaking concentration to get your action back.
5) (Bonus) Less menu clutter. There isn't _both_ a 'Stop Concentration' and an 'Implode' item, so you can't accidentally click on the former when you were aiming for the latter.

So, it seemed like the best way to handle a concentration-allowing-an-action spell was to make the condition that allows the action also track the concentration.

One thing just occurred to me, though. Maybe it'd be better to factor out the concentration-handling stuff into a separate 'active concentration' condition, and have the implosion condition signal it. That way the hooks wouldn't have to be replicated by each spell. The above argument is the same for not using the existing `sp-Concentrating` effect, though.

DudeMcDude:
About 4.c. This reminds me of stuff like the defensive attack feat (i forgot the name) where you could try to cheat it by switching it on after attacking. So the game keeps track of what you actually did in the arg variables, and directly prevented that sort of abuse. You could do the same here: send some signal that indicates you used up your concentration, thus disabling the option to get your standard action back. Normally I'd assume this requires adding an arg, and thus a no-no, but it looks like there are 2 unused args for sp-Concentrating, so perhaps the extension method is warranted.

The other option is to add an auxiliary condition, which might as well include the whole thing I suppose. The only caveat I can think of in that regard is whether the game somehow assumes that the only concentration condition is sp-Concentration, i.e. directly checking it somewhere, but that doesn't seem to be the case.

Not sure which option is preferable yet.

dolio:
It might be possible to implement the whole thing by extending the existing concentration condition. My concern was that I'm not sure what all the arguments are for. 0 is spell id of course, but it has two more. It doesn't have a begin round hook, so 1 isn't duration, I guess. It might just be that so many generic hooks are used, it needs to have 3 arguments to avoid crashing. But I don't know what side effects storing things in those other arguments would cause.

The only concern I see with a second concentration condition is that the original responds to Q_Critter_Has_Condition for sp-Concentrating. However, I don't know whether that's checked, because there is also Q_Critter_Is_Concentrating which is less of a worry to implement.

dolio:
Oh, another side note:

Some of the way `sp-Concentrating` works is a bit weird. They seem to have weird interpretations of the targeted-by-a-spell rules.

1) Damaging spells don't cause concentration checks based on the damage. Instead...
2) _All_ spells targeting the concentrating party force concentration checks corresponding to what the spell's save DC would be.

This is almost certainly worse for damage spells. It's kind of a a 'meme' that readying a scorching ray is far superior to readying a counterspell (although I'm not sure if the while-casting check is the same as sp-Concentration). But these rules make it so that instead of DC 10+(4*R)d6+N to maintain concentration on a level N spell, it's 13+N.

But also, it means that e.g. curing someone who is concentrating on a spell requires them to make a concentration check.

In P&P, I think certain non-damaging spells are indicated as distracting people affected by them (the example in the SRD is Storm of Vengeance), and those are the only ones that use the spell DC rules. I'm not sure how many offensive-but-non-damaging spells might not specify one way or the other, though.

PR #557, opened 2021-07-30
Spell Heart Ripper from the SC

Sagenlicht:
@dolio 
This spell contains a modified spell_utils. I often need to check if a target has a certain mc_type, and I finally added the check to the spell_utils.


DudeMcDude:
Is this still a WIP? It looks finished...

Sagenlicht:
Is there a better way to handle the XP issue? Other than that the spell is pretty much done and I tested it a fair amount already.

DudeMcDude:
Ah ok cool.
I think in temple+ you can add an arg to the "kill by effect" call that denotes who the attacker is, so you don't have to apply manual damage.

Sagenlicht:
Ah ty for pointing this out to me. Fixed

DudeMcDude:
> Ah ty for pointing this out to me. Fixed

Yeah I added it for the Assassin :)

BTW there's a bunch of user bug reports on the Spell Compendium thread, can you look into it?

PR #558, opened 2021-07-30
Stricter Concentrating status

dolio:
I noticed you can extend existing conditions from Python, so I put together a suggestion of how to make concentration spells more accurate. The idea is that at the beginning of your turn, `sp-Concentration` removes your standard action, and if you choose the `Stop Concentration` radial, it gives it back.

There are (at least) two problems with this. The first is that `EK_D20A_STOP_CONCENTRATION` never gets dispatched. The second is that setting the `hourglass_state` in the manner I'm doing it doesn't have any effect on the real state.

I verified the latter by trying out a more complicated scheme involving a second condition that gets added by `sp-Concentration`. However, it seems like it'd be better to just be able to hook into the action perform if possible. It actually looks like that could solve both problems, even, because `Dispatch37` can carry a `tb_status`. However, it looks like it's always null in the original code.

### DIFF
--- a/tpdatasrc/tpgamefiles/scr/tpModifiers/sp_concentration.py
+++ b/tpdatasrc/tpgamefiles/scr/tpModifiers/sp_concentration.py
@@ -1,0 +1,27 @@
+from templeplus.pymod import PythonModifier
+from toee import *
+import tpdp
+import tpactions
+
+def UseAct(char, args, evt_obj):
+	if tpdp.config_get_bool("StricterRulesEnforcement"):
+		if evt_obj.tb_status.hourglass_state > 1:
+			# use up a standard action
+			evt_obj.tb_status.hourglass_state = 1
+	return 0
+
+def GiveAct(char, args, evt_obj):
+	if tpdp.config_get_bool("StricterRulesEnforcement"):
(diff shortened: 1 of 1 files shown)


DudeMcDude:
I might have missed it, but where does it say concentration consumes a standard action?

dolio:
It's here:

  https://www.d20srd.org/srd/magicOverview/spellDescriptions.htm#durationConcentration

There are maybe a couple ways to get around it in some splat books.

Sagenlicht:
This could be a nice usage for the new popup window. Do a popup at start of round and ask if you want to concentrate on the spell. If not send a stop concentrating signal else substract the standard action.

This would also visual the changes in enforcing the concentration rule.

dolio:
I kind of figured that having to actively do something each round would be annoying. It seems like defaulting to maintaining concentration makes sense as long as the cost of that can be rolled back correctly.

dolio:
I did a little searching, and I think I know why `get_cur_seq` behaves the way it does. The lambda signature is `[]() -> ActnSeq&`, but the pybind automatic return policy is making a copy of the result because it's a struct, even though it's a reference result. So modifying the result on the python side has no effect on the C++ structure.

PR #572, opened 2021-08-07
Fixed discrepancy between MoveSequenceParse pathfinding query and Can…

DudeMcDude:
…PathTo PF utility


DudeMcDude:
This resulted in a loop that a user has reported.
The AI used CanPathTo to determine whether to enter combat with the party, but then the AI itself uses FindPath with a different setting and thus failed to execute. Normally this wouldn't be a problem, except in the case of a single hidden AI critter it causes combat to immediately end, and thus trigger a loop.
(situation was encountered with Crayfish in Moathouse, where party was approaching with a tight formation such that the targeted PC was deemed surrounded and thus not a valid PF target - this may be a bug in itself however)

PR #573, opened 2021-08-07
Mirror Image error due to changing get_metamagic_data return type

DudeMcDude:
Heads up: @anatoliy-savchak @dolio @doug1234 @Sagenlicht 
I don't know if you've used this anywhere, but the spell_packet method get_metamagic_data() has been changed from returning integer to tpdp.MetaMagicData object. (This borked the Mirror Image script, which was just caught by a playtester)


Sagenlicht:
The deal_spell_damage  function in damage_critter.py uses this.

DudeMcDude:
I know, that's why I changed the API in the first place ;)

doug1234:
Thanks for the heads up.  I'll look through some other scripts and see if it might be causing problems.  

BTW:  Is there still a way to get the metamagic data as an integer in python?

DudeMcDude:
No. Though I'd rather the python side dealt with the object rather than expose the internals of how the bytes are packed, in case we ever change it.

PR #581, opened 2021-08-15
Add get_d20_spell_data to EventObjD20Query

dolio:
Trying out something simple on the C++ end.

`Q_SpellInterrupted` sends a pointer to the `D20SpellData` for the spell being cast in the `data1` field, but there was no way to get access to the information (I think).


dolio:
Seems to work. I tested locally with an extension of Entangle that does the proper concentration check when casting a spell. The callback in the original just doesn't do anything.

PR #590, opened 2021-08-22
Obscuring Mist fix

dolio:
This adds accurate particle effects for Obscuring Mist.

The existing effects are about 5ft in radius, but the spell creates a 20ft radius effect. I'm not sure how the merging works for this file, so it might not override the old particles. However, if it isn't, it doesn't look to bad to me.

I have another suggestion for the spell which I can add to the PR. The rules file could be changed so that the targeting acts more like Bless. You can only target yourself, but you get to see the radius before the spell is cast. Right now it immediately casts as soon as you select it from the radial, because it's set up as a personal spell.

I wasn't sure what other people would think of the latter change, though, so I haven't included it yet.


DudeMcDude:
I sounds like it'd be better if all personal + range spells behaves this way at the engine level (requiring click to comfirm) rather than manually changing each one. As a bonus, it'd display the radius without having to mouse your character. 
Fortunately the picker system is largely replaced already so it shouldn't be too hard to do.

dolio:
Oh. That'd be really nice, actually.

PR #594, opened 2021-08-26
Fix some mirror image logic

dolio:
- If the attack roll against the image missed, the attacker was instead getting an attack against the real target, rather than just missing
- The code was treating the images as having 20 AC, so attack rolls against images would almost certainly miss too often
- Removed the DC roll log message, so it's just shown as the attack roll that it is (the DC message was replacing the attack roll)
- Removed some now unnecessary arguments

Note: Mage Armor erroneously contributes to image AC. I think this is because it applies to all touch attacks. It's actually only supposed to apply against incorporeal attacks that otherwise bypass armor. That's a separate issue, though.

### DIFF
--- a/tpdatasrc/tpgamefiles/rules/d20_combat/to_hit_processing.py
+++ b/tpdatasrc/tpgamefiles/rules/d20_combat/to_hit_processing.py
@@ -32,7 +32,7 @@ def add_percent_chance_history_stub():
     return
 
 
-def mirror_image_attack_roll(d20a, spell_id):
+def mirror_image_attack_roll(d20a):
     performer = d20a.performer
     target = d20a.target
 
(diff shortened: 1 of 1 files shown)


DudeMcDude:
> It shouldn't be 10 either. The spell says 10+dex+size. I figured it was dispatching to account for someone being denied their dex bonus. And I figured it was using OnGetAC because that actually excludes most bonuses.
> 
> Are there other bonuses that get included in that call? Because Mage Armor being included for all touch attacks is just incorrect itself.

As far as I can tell OnGetAC includes all bonuses except:
Invisibility
Feint
Magic Circle (Inward)

Of course, touch attack nullifies most bonuses.
There may also be malluses that (erroneously?) affect it.

PR #597, opened 2021-08-28
Calculate cleric domain slots properly

### LINKED ISSUE #596: Practiced Spellcaster gives domain slots
E.G. a Wizard 4/Cleric 1 with Practiced Spellcaster - Divine will have 3rd level domain slots. They will only know 1st level domain spells, but the slots appear early.

Seems to be particular to domain spells. Normal cleric slots do not appear early.

dolio:
Was using base caster level. Should use effective cleric level.

Should fix #596


dolio:
Actually, this does not fix #596. I misunderstood the function name, and it was actually calculating the domain spells _known_ incorrectly. But I guess it wasn't used.

Still looking for where the domain slots are calculated.

dolio:
Okay, that second one was the culprit.

PR #598, opened 2021-08-29
Countersong

dolio:
Working on countersong. Pretty sure it didn't actually do anything before. The `OnCountersongSaveThrow` handler checks an argument for the replacement saving throw, but it was always set to 0.

There are other issues as well, but without fixing that part first, nothing else can work.

One question I have is: the way Countersong is described is that you can use the bard's Perform check instead of a saving throw. The way the _code_ is written is to use the perform check to construct a bonus to the save, and still fail on a 1. Skill checks don't automatically fail on a 1, though. So, should I leave it as-is; a variable save bonus? Or should I modify it to act like it's actually the bard countering with a skill roll?

I don't think preserving vanilla behavior is really a factor, because the 0 argument thing means that it would never have done anything anyway (no DC 0 checks), so the existing behavior is hypothetical only.

### DIFF
--- a/TemplePlus/condition.cpp
+++ b/TemplePlus/condition.cpp
@@ -6373,6 +6373,7 @@ int ClassAbilityCallbacks::BardicMusicBeginRound(DispatcherCallbackArgs args){
 		if (!gameSystems->GetObj().IsValidHandle(tgt)) {
 			tgt = objHndl::null;
 		}
+		auto rollResult = 0;
 		switch (bmType)
 		{
 		case BM_INSPIRE_GREATNESS:
@@ -6385,8 +6386,7 @@ int ClassAbilityCallbacks::BardicMusicBeginRound(DispatcherCallbackArgs args){
 			party.ApplyConditionAround(args.objHndCaller, 30, "Inspired_Courage", objHndl::null);
 			return 0;
 		case BM_COUNTER_SONG: 
-			auto rollResult = 0;
-			skillSys.SkillRoll(performer, SkillEnum::skill_perform, 0, &rollResult, 1);
+			skillSys.SkillRoll(args.objHndCaller, SkillEnum::skill_perform, 0, &rollResult, 1);
 			party.ApplyConditionAroundWithArgs(args.objHndCaller, 30, "Countersong", {0, rollResult, 0});
 			return 0;
 		case BM_FASCINATE: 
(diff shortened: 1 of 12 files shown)


dolio:
So, here's a conundrum.

Countersong should presumably have a possibility of ending captivation from harpies. I can extend the `Captivating Song` condition to end via countersong like a normal spell. However, it looks like that condition just adds a second condition `Captivated`, which does not reference the original spell. So e.g. there's no good way to look up the DC of the effect.

Should I just assume that all `Captivated` conditions come from harpies? Or are there other monsters that use it?

DudeMcDude:
Oh man, it looks like I wrote that before I knew what I was doing. Why write the singer ID if you take it from the spell caster anyway??

Anyway it's safe to assume it's the harpy song, yes. Though it looks like you could add the spell id to the args, they seem unused.

dolio:
When I look in the DM tool, it seems like all the arguments to `Captivated` get populated. It's not obvious how from the C++, though. Anyhow, I wasn't sure it was wise to mess with them.

DudeMcDude:
Oh right, args 2-7 are taken by the singer object ID.
Arg1 is the particle system ID.
Arg0 is duration.

It could use a rewrite, all this data (except duration) should be stored inside the spell.
I had also meant for the singer ID to be used to compel the Captivated character to approach, but never got around to it (I guess I postponed it until I figured out the AI system).

dolio:
I think this might be ready to go.

I was trying to figure out if I needed to do something with Suggestion. I see that there are files for it, but I can't actually get it as a spell. Is that the effect that works nothing like the P&P spell, and is more like Dominate Monster?

DudeMcDude:
Are you talking about the spell or bard ability?
The latter is an ability I added, which I interpreted as Cause Fear rather than Dominate/Charm :)

dolio:
No, there's files for a Suggestion spell in the Co8 mod, I think. I'm not clear if it was in the original game or if the condition was discovered and someone wrote the spell to use it, but if it's the one I'm thinking of, it's way too powerful for the level 3 spell that it is.

I guess I'll look at the bard suggestion, too. They're a little tricky to remove after the fact, though.

DudeMcDude:
Ah yes, the original Suggestion was was basically Dominate Monster and was converted by Co8 to just that.

dolio:
Okay, bardic suggestion can now be removed by countersong appropriately. I also took the liberty of adjusting the caster level on fascinate, because it was pulling in practiced spellcaster, which gives you extra targets.

Issue #585, opened 2021-08-20
Prefer full attack + Stunning Fist


dolio:
It seems that if you have the game set to prefer full attacks, then stunning fist will automatically initiate a full attack if possible. Most other special attacks don't do this; they're usable within a full attack, but selecting them without first initiating a full attack will use them as a single attack.

This wouldn't be a big problem as a default, but there doesn't seem to be a way to _not_ do the full attack.

doug1234:
I added the prefer full attack feature.  I can look into this one.

DudeMcDude:
Is this issue only applicable to Stunning Fist?

DudeMcDude:
Err, are you sure it's related to that setting? I just tested and it did the same with prefer 5' step turned off.

dolio:
Hmm, I don't actually remember if I tested with it off. It's possible I just presumed that the setting was meaningful.

DudeMcDude:
Nope, it isn't. Also it's a vanilla ToEE bug (just verified).

dolio:
Hah. Okay then. Sorry for the wild goose chase.

DudeMcDude:
Sure, why not.

dolio:
Interestingly, on further study, the ticker for stun is trying to do the right thing. It stores the monk's initiative in one of the arguments, and then on each X->Y change in initiative, it does some checks. But the case where the initiative wraps around from a low number to a high number must be faulty.

PR #603, opened 2021-09-03
Fix recently found bugs

doug1234:


TemplePlus/action_sequence.cpp
```diff
@@ -1227,7 +1227,10 @@ uint32_t ActionSequenceSystem::MoveSequenceParse(D20Actn* d20aIn, ActnSeq* actSe
 						}
 					} 
 					else {
-						if (!tbStatCopy.hourglassState || 	tbStatCopy.hourglassState == 4 && (config.preferUse5FootStep || !objects.IsPlayerControlled(d20a->d20APerformer))) { // added for AI to take 5' steps when it still has full round action to exploit
+						// Added for AI to take 5' steps when it still has full round action to exploit.  Players can select an option for them as well.  
+						// Don't use the player option for stunning fist since there wouldn't be a way to use it without the 5 foot step and full attack.
+						if (!tbStatCopy.hourglassState || 	tbStatCopy.hourglassState == 4 && ((config.preferUse5FootStep && (d20->globD20Action->d20ActType != D20A_STUNNING_FIST))
```
DudeMcDude, line 1232: I prefer a generic fix rather than ad-hocing specific actions (unless Stunning Fist is a real exception).

tpdatasrc/tpgamefiles/scr/tpModifiers/brutal_throw.py
```diff
@@ -6,9 +6,20 @@
 
 print "Registering Brutal Throw"
 
+#Check if the weapon can only be thrown
+def IsThrownOnly(wpnType):
```
DudeMcDude, line 10: Have you found an equivalent function in the engine? Or is it likewise hardcoded?

TemplePlus/action_sequence.cpp
doug1234: For some reason stunning fist works differently than everything else according to issue #585.  The best way to fix the issue might be to rework the stunning fist radial menu.  Not sure we want to change that around though.

tpdatasrc/tpgamefiles/scr/tpModifiers/brutal_throw.py
doug1234: As far as I can tell there is no function like this in the engine.  In fact the one place I can think of where it should be used on the C++ side is done incorrectly (same bug as my feat).  I will add a version in C++, fix the bug and make the C++ version accessible in python.

DudeMcDude: Wait, so is Javelin actually throw only in the game, or just by the rules? (btw technically you could use it for melee, but with a penalty)

TemplePlus/action_sequence.cpp
DudeMcDude: Also it doesn't seem logical that the place where it should be fixed is inside MoveSequenceParse.
I'll look into this.

tpdatasrc/tpgamefiles/scr/tpModifiers/brutal_throw.py
doug1234: The javelin is thrown only in game.  I can't see any way to make a melee attack with it.

TemplePlus/action_sequence.cpp
doug1234: It might be better to do this further up stream.  Not sure on that.  Other than that I think the pull request is ready.

doug1234: After reading the comments on the issue, should I revert this change and merge whats in there?

DudeMcDude: Yes please.

PR #604, opened 2021-09-04
Fixed Faith Healing

Sagenlicht:


Sagenlicht:
This SC Fix includes Faith Healing corrected to 8+level (max 5) and crashing issues from Conviction and Divine Protection

PR #605, opened 2021-09-06
Greater Mirror Image

dolio:
I believe this is all the files necessary for Greater Mirror Image. Tested a bit locally and seems to work.

I could add a few more beguiler spells if desired. Or it could wait. Either way.


DudeMcDude:
Hmm, does it address multiple castings of the same spell? Doesn't seem so. Or overlap with normal MI for that matter.

dolio:
You're right.

Here's a question, though. What should I have it do?

Normal mirror image just cancels the previous spell. However, all the machinery is in place to have them properly overlap. If you cast multiple copies, you get the benefit of whichever spell has the most images currently.

Should I try to make regular Mirror Image allow overlapping in this way? Or make the new one work the same as the existing one? Or not worry about them being exactly consistent right now?

DudeMcDude:
I think the way it should work is to return the greatest value (i.e. max of event_obj.return_val and arg2), and simultaneously deduct images from all active MIs.

dolio:
If a hit is going to deduct from every spell, then I might as well just implement overwriting with new spells. Having two spells active at the same time is not going to be worth while enough for anyone to ever do it, and it's easier to implement (since it's how mirror image already works).

DudeMcDude:
Fair enough.

PR #617, opened 2021-09-16
Fix sudden spells and abj camp for spontaneous casters.

### LINKED ISSUE #614: Abjurant Champion meta-magic effect makes casting time full round

doug1234:
closes #614


DudeMcDude:
Does this affect rapid metagic?

doug1234:
I don't see how it would, it just won't count meta magic effects added through python when seeing if a spontaneous spell should be upped to full round.  I can double check if you think it will.

DudeMcDude:
Oh right, the action cost modify comes after that. Should be ok.

Issue #542, opened 2021-07-17
Bonus Spell Slots


dolio:
This came up in a PR discussion.

It'd be useful to be able to have conditions that provide bonus spell slots. Things that would use that are, at least, Rary's Mnemonic Enhancer and Precocious Apprentice.

In browsing the original code, it looks like one of the dispatchers might have been intended to support this (my recollection was maybe 57 or 58). However, if that was the case, it got dropped.

dolio:
Noting: Dragon Disciple would use this.

DudeMcDude:
> Noting: Dragon Disciple would use this.

 Very prescient 😆

DudeMcDude:
So assuming we add a dispatch for this:
1. What info should be sent out in the event object? I'm guessing class enum is enough.
2. I'm thinking the event object should contain a list of bonus slots per level, and the recipients should be able to increment (or decrement, for stuff that requires sacrificing slots e.g. archmage). 

Thoughts?
Pinging @Sagenlicht @doug1234 @anatoliy-savchak @shartte

dolio:
Another use case is rings of wizardry.

What those do is double your inherent number of slots for a given level. However, it _only_ doubles the base number. So, a 3rd level specialist wizard with a ring of wizardry 1 would get 5 extra 1st level slots. They wouldn't get 6, because the specialist slot is not doubled, and also the bonus slots from high intelligence also aren't doubled.

So, the callback needs to be able to tell how many (effective) levels of the relevant class the character has. I guess that's easily available with just the class enum, though.

Sagenlicht:
> So assuming we add a dispatch for this:
> 
> 1. What info should be sent out in the event object? I'm guessing class enum is enough.
> 2. I'm thinking the event object should contain a list of bonus slots per level, and the recipients should be able to increment (or decrement, for stuff that requires sacrificing slots e.g. archmage).
> 
> Thoughts?
> Pinging @Sagenlicht @doug1234 @anatoliy-savchak @shartte

Sounds great, don't think more is needed.

anatoliy-savchak:
Looks good

PR #619, opened 2021-09-16
Dragon Discipile WIP

Sagenlicht:
Created a Dragon Discipile Help File (WIP)
Created Dragon Discipile Feat text files (Heritage incomplete)
Started to work on Dragon Discipile Feats


DudeMcDude:
Good start!
I think I'd unify all the heritages to one and make the effect depend on arg.
To get around the general issue of feature selection, for now you can have the player choose the heritage type via radial menu.
Also note the addmesh thing I added - I had the wings feature in mind 😉 It won't actually grant flight but maybe we could have it give faster movement speed.

Sagenlicht:
> Good start!
> I think I'd unify all the heritages to one and make the effect depend on arg.
> To get around the general issue of feature selection, for now you can have the player choose the heritage type via radial menu.
> Also note the addmesh thing I added - I had the wings feature in mind 😉 It won't actually grant flight but maybe we could have it give faster movement speed.
> 
> Oh and blindsense is kinda redundant in this game I think. I suppose it could he integrated in spot checks, but meh.

My plan was to store the Heritage in the Breath Weapon and in the Dragon Apotheosis Functions as an arg anyways, those are the only functions that care about the heritage. Saving the heritage as an arg makes it easier to add new dragon types in the future if really needed. Now I really like the idea to handle the heritage via a radial menu. This saves alot of feat headaches indeed :) Will do it that way.

DudeMcDude:
BTW, maybe the dragon heritage should have its own condition - for example if any other classes use that kind of stuff (e.g. Pathfinder Sorcerers).

Sagenlicht:
Updated Help File (close to finished) and fixed typo
Added Placeholder Breath Weapon Cone particles
Added Breath Weapon Radial and Action
Added Spell Dragon Disciple Cone Breath

@DudeMcDude 
https://github.com/GrognardsFromHell/TemplePlus/blob/03126730533a0e19c4a1168bfda7968b7549e204/tpdatasrc/tpgamefiles/scr/tpModifiers/dragon_disciple.py#L232-L239
The breath weapon runs into an error, though I am pretty sure it should work. I either have a brain afk or something is broken as I am pretty sure this did work in the past.

Sagenlicht:
Added Ability Boost Feature.

Known Bugs:
Select Heritage Radial does not work as intended and chooses wrong Heritage
Breath Weapon is not executed

Missing:
Claws and Bite Attacks
Blindsense

Sagenlicht:
Added Cone Breath Weapon particles
Added Line Breath Spell
Added Line Breath particles for Acid

DudeMcDude:
I think the claw/bite stuff needs engine level hooks. IIRC it looks up obj_f fields that are NPC only.

tpdatasrc/tpgamefiles/scr/tpModifiers/dragon_disciple.py
```diff
@@ -132,15 +141,215 @@ def OnLevelupSpellsFinalize(attachee, args, evt_obj):
...
+        return 0
+
+def performBreathWeapon(attachee, args, evt_obj):
+    chargesLeft = args.get_arg(0)
+    currentSequence = tpactions.get_cur_seq()
+    spellPacket = currentSequence.spell_packet
+    newSpellId = tpactions.get_new_spell_id()
+    spellPacket.caster_level = attachee.stat_level_get(classEnum)
+    tpactions.register_spell_cast(spellPacket, newSpellId)
+    tpactions.trigger_spell_effect(newSpellId)
```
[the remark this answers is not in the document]
Sagenlicht: Will do, was actually surprised the caster doesn't turn automatically when I use target mode: ray. The caster e.g. turns automatically when you use target_mode: cone

DudeMcDude:
> > I think the claw/bite stuff needs engine level hooks. IIRC it looks up obj_f fields that are NPC only.
> 
> Ok

Oh, on that note, are there other things that grant natural attacks? I know polymorph does, but it does so through mimicking an existing proto entry.

anatoliy-savchak:
Wow, looks great!

Question - why have you decided to go with Breath Weapon as a spell implementation? 
For example I did similar here: https://github.com/anatoliy-savchak/ToEE.ZMOD.ShatteredGatesOfSlaughtergarde/blob/master/modules/zmod_sgos_core/scr/tpModifiers/line_of_acid.py

But I feared that Supernatural ability should not be treated as spell. So no spell resistance nor AOO nor counterspelled nor dispelled.

anatoliy-savchak:
> Unless I'm missing something, the above is for AI with automatic target selection, which is not suitable for PC class obviously. Making this an action with a spell makes use of the UI picker infrastructure.

Sure. Thanks Sitra!

Look, my comment is not criticism per se. I saw this and now will try to accommodate such technique in my modules :)
Great work @Sagenlicht ! :)

DudeMcDude:
> > Unless I'm missing something, the above is for AI with automatic target selection, which is not suitable for PC class obviously. Making this an action with a spell makes use of the UI picker infrastructure.
> 
> Sure. Thanks Sitra!
> 
> Look, my comment is not criticism per se. I saw this and now will try to accommodate such technique in my modules :)
> Great work @Sagenlicht ! :)

I don't know when you wrote that, but actually ai tactic "cast area" was broken until quite recently. I'm still not sure how well it works for cones actually.

Sagenlicht:
Updated help file. Help file is finished and only needs to be corrected for things that maybe gets change due to technical limitations (see below).

About the bite/claw thing. A few spells in the Spell Compendium would add claw/bite attacks as well. I am not aware of any other claw/bite attacks outside spells or non humanoid races as sources for such attacks. Claw attacks are usually two-weapon attacks and the penalty for it is usually reduced by the (in ToEE I belive non existing) [https://www.d20srd.org/srd/monsterFeats.htm#multiattack](url) feat.

Blindsense
What is your prefered handling @DudeMcDude for the Blindsense feat? Ignore it or give a spot bonus? I don't think the feat does much more. If a spot bonus is of no real relevance, I can simply drop the feat.

Wings
I added a feat for the wings, so the addmesh thing can be hooked.

Breath Weapon
All particle effects are done, only fire smokes needs to be blackend.
I am bit stuck on the facing for the breath weapon, besides that everything is working fine.

Spell Selection
I did reread the spell advancement of the Dragon Disciple and it is not as I thought it would be. Instead of gaining a simple advancement in the highest arcane casting class (which I assumed) it only gains bonus spells for an existing spell level, which can always newly be selected every time its applicable.
I am unsure if this is actually doable atm in ToEE?
[https://www.d20srd.org/srd/prestigeClasses/dragonDisciple.htm](url)

@anatoliy-savchak  ty :)

anatoliy-savchak:
> I am not aware of any other claw/bite attacks outside spells or non humanoid races as sources for such attacks.

Many monsters have these natural attacks. But unfortunately you cannot mix Natural Attack and Weapon Attack in ToEE\T+ ATM.
image

DudeMcDude:
> About the bite/claw thing. A few spells in the Spell Compendium would add claw/bite attacks as well. I am not aware of any other claw/bite attacks outside spells or non humanoid races as sources for such attacks. Claw attacks are usually two-weapon attacks and the penalty for it is usually reduced by the (in ToEE I belive non existing) [https://www.d20srd.org/srd/monsterFeats.htm#multiattack](url) feat.
Well I was wondering if it should be converted to a general thing rather than DD specific. I guess it can always be changed later though.  
> 
> Blindsense
> What is your prefered handling @DudeMcDude for the Blindsense feat? Ignore it or give a spot bonus? I don't think the feat does much more. If a spot bonus is of no real relevance, I can simply drop the feat.  

Yeah I think we can take a pass on that.  

> 
> 
> Spell Selection
> I did reread the spell advancement of the Dragon Disciple and it is not as I thought it would be. Instead of gaining a simple advancement in the highest arcane casting class (which I assumed) it only gains bonus spells for an existing spell level, which can always newly be selected every time its applicable.
> I am unsure if this is actually doable atm in ToEE?
> [https://www.d20srd.org/srd/prestigeClasses/dragonDisciple.htm](url)
> 
I don't understand, is this different than Eldritch Knight?

Sagenlicht:
> I don't understand, is this different than Eldritch Knight?

Yes completely.
EK, as almost all PrC's gets normal spell progression. When EK gains a level it counts as getting a level in the "original" class. This means an EK 4 / Sor 1 can cast 2nd level sorcerer spells.

An Dragon Disciple only gets bonus spells no progression or even spells per day. A DD 10 / Sor 1 still could only cast 1st level spells only three per day.

All the DD gets is more spells known and in by raw he could choose every time he gets a bonus spell in which spell level he wants to have the new spell.
Example: A 5 Level Sor can cast 1st and 2nd level spells. When she gains the first level of DD she still can't cast level 3 spells (an EK would grant access to those) but she can choose to add either a 1st level or 2nd level spell to her known spells.
In addition to this, if character would be 2bard/4sorc and then gains DD levels she always could by raw choose to add a new level 1 bard spell or a new Sorc spell either 1st or 2nd level. I guess the 2nd example is more of an edge case but in theory it work like that.

dolio:
The Dragon Disciple bonus spells aren't known spells. They are per-day spell slots. You get to add one spell-per-day to any level you can already cast on those levels.

I think there's no way to give them in ToEE. The engine just hard-codes bonus spell slots from high ability scores as the only possibility. I made issue #542 a while back about it.

Sagenlicht:
Ah ty for correcting this @dolio. Ok they are not more known spells but more spells per day then. But yeah it is still not the normal PrC spell progression sadly :(

dolio:
Yeah, it's surprisingly weak, since it doesn't seem like the rest of the class is really sufficient to make you a competent fighter.

Incidentally, I've noticed that the original (I think) designers misinterpreted Multiattack, too, since you mentioned it. What it's supposed to do is make 'secondary' natural weapon attacks hit at -2 (rather than -5). However, what they did instead was give monsters with Multiattack a bonus attack, and this ends up being an unarmed strike (dealing nonlethal damage) rather than an actual natural attack. I think the way natural attacks are specified (currently) just lets you pick arbitrary to-hit penalties, so they should have just picked -2 for creatures with the feat, and -5 for creatures without.

Two claws _could_ be the primary natural attack, I think. They just happen not to be in the case of DD.

Sagenlicht:
Fixed Breath Weapon Fire particle smoke
Dropped Blindsense

Bugs:
Performer doesn't rotate when using Breath Weapon and is not facing the target(s)*

Missing
Claws and Bite attack
Wings

*I can't get attachee.turn_towards(target) to work properly :( I do not have access to any other appropriate anim_goals I think. In the code I used anim_goal_push_attack as a placeholder for now. Though turn_towards isn't the best solution for a cone spell anyways.

As the DD normally would get fly and this can't be done in ToEE, we could offer something else in return. 

@DudeMcDude already thought about increasing speed, to throw another idea in the ring how about adding a modification to the Breath Weapon, depening on the type of element:
* Acid and Fire add damage over time for 1d4 rounds on a failed reflex save (oozing or burning)
* Cold slows on a failed fortitude save for 1d4 rounds
* Electricity stuns (or dazes if stun is too strong) on a failed fortitude save for 1 round

If we would add other element types, this could be expanded as well (e.g. negative adds a temp neg level for 1d4 turns on a failed fort save, sonic deafens etc).

DudeMcDude:
Oh sorry I forgot about this.
Increasing speed - for some reason I thought all flyers have higher speed. Probably from looking at the fly spell. But that's not the case here. 
I could think of adding a "Flying" condition, that could be turned on/off to avoid stuff like grease/entangle etc, and make melee attacks impossible when outdoors, but honestly the utility seems very limited for what will likely be a combat frontliner.

Is the turn towards attacker and missing bite/claw the only remaining issues?

Sagenlicht:
Oh I like your idea about the flying condition actually. You could also disable AoO when moving/charging with this condition I guess, which could be quite relevant. I don't know if it's possible to let the character levitate above its space for the time he/she flies but everything else seems actually quite reasonable. 

Missing features are:
* Claws and Bite attacks
* Proper Facing for the Breath Weapon
* Bonus Spells Slots (Ring of Wizardry Issue)

anatoliy-savchak:
>I could think of adding a "Flying" condition, that could be turned on/off to avoid stuff like grease/entangle etc, and make melee attacks impossible when outdoors, but honestly the utility seems very limited for what will likely be a combat frontliner.

Personally I agree about the Grease. Good point. Not sure about entangle, probably fine. This is complicated subject.

I **do** recommend bringing this topic to co8 forum and brainstorm it.

TemplePlus/python/python_integration_d20_action.cpp
```diff
@@ -133,9 +133,9 @@ PYBIND11_EMBEDDED_MODULE(tpactions, m) {
 		return TRUE;
 	});
 
-	m.def("get_cur_seq", []()->ActnSeq &{
-		return **actSeqSys.actSeqCur;
-	});
+	m.def("get_cur_seq", []()->ActnSeq *{
+		return *actSeqSys.actSeqCur;
+	}, py::return_value_policy::reference );
```
DudeMcDude, line 138: I've changed get_cur_seq to return a reference rather than a copy. I vaguely recall someone needed this change, pinging @dolio @Sagenlicht @anatoliy-savchak @doug1234

Sagenlicht: It wasn't me, though I do use get_cur_seq in quite a few python actions.

DudeMcDude: > Oh I like your idea about the flying condition actually. You could also disable AoO when moving/charging with this condition I guess, which could be quite relevant. I don't know if it's possible to let the character levitate above its space for the time he/she flies but everything else seems actually quite reasonable.
> 
> Missing features are:
> 
> * Claws and Bite attacks
> * Proper Facing for the Breath Weapon
> * Bonus Spells Slots (Ring of Wizardry Issue)

As you might recall I've added a hook for bonus spell slots; are you able to make use of that?

Sagenlicht: I am sure I can :)

Thanks alot for the new anim_goal, can't wait to use them for a lot of different things :)

I did  postpone my work on the DD due to some Breath Weapon issues, will resume with the DD tomorrow and will try to add the bonus spells.

Sagenlicht:
I switched the DD to use the Draconic Heritage Feat instead the radial workaround for handling the heritage.

This push also includes the heritage_feat_utils.py, which is required for interacting with heritage feats, and the DD now uses the Draconic Heritage Feat. I think this is the way to go, as it generalizes the heritage handling and was the reason why I started to work on all the heritage feats (I've done the Fey Line completly and alot of the Draconic are also already done), though I do not want to include them in the DD branch, will push them afterwards.

I also created a BreathWeaponModifier class to handle all upkeep work for breath weapons (charges, cooldown), as the Dragon Shaman also needs this (also to generalize)

Lastly this push also includes the Extra Exhalation Feat, so a DD can increase his Breath Weapon Daily uses. Have done another Breath Weapon Feat (Exaled Immunity), won't be included in the DD branch, will get a seperate push.

Will work on the bonus spells tomorrow, afterwards only the claw/bite thing is missing and the DD doesn't use any workarounds anymore :)

DudeMcDude:
Making explicit bonus spell feats for every class and spell level is pretty icky... why go down that route?

Sagenlicht:
> Making explicit bonus spell feats for every class and spell level is pretty icky... why go down that route?

Oh I am unsure how to handle this properly, this was my first draft idea. 

The Dragon Disciple could get a bonus spell for every level and spell class he chooses as he gains level. It is not limited to his highest class, arcane or whatsoever. So in theory a 2nd Level Cleric/2nd Level Sorc/1st Level Bard that then goes Dragon Disciple could choose to add a bonus spell to: Cleric Spell LEvel 0 or 1, Sorc Level 0 or 1 or to his Bard spell Level 0 every time he gets a bonus spell slot.

Now I am unsure how this Bonus Spell thing is needed by other classes or items like a spell storing ring, so my first idea was let's do some conditions everyone could use. But I am open for suggestions :)

Possible bug: I don't get evt_obj.caster_class to work, it runs into an error.
EventObjSpellsPerDay "spell_level", "base_caster_level" and "bonus_list" work fine.

DudeMcDude:
> I switched the DD to use the Draconic Heritage Feat instead the radial workaround for handling the heritage.
> 
> This push also includes the heritage_feat_utils.py, which is required for interacting with heritage feats, and the DD now uses the Draconic Heritage Feat. I think this is the way to go, as it generalizes the heritage handling and was the reason why I started to work on all the heritage feats (I've done the Fey Line completly and alot of the Draconic are also already done), though I do not want to include them in the DD branch, will push them afterwards.
> 
> I also created a BreathWeaponModifier class to handle all upkeep work for breath weapons (charges, cooldown), as the Dragon Shaman also needs this (also to generalize)
> 
> Lastly this push also includes the Extra Exhalation Feat, so a DD can increase his Breath Weapon Daily uses. Have done another Breath Weapon Feat (Exaled Immunity), won't be included in the DD branch, will get a seperate push.
> 
> Will work on the bonus spells tomorrow, afterwards only the claw/bite thing is missing and the DD doesn't use any workarounds anymore :)

Excellent!

tpdatasrc/tpgamefiles/scr/tpModifiers/draconic_heritage.py
```diff
@@ -0,0 +1,78 @@
...
+        evt_obj.return_val = 1
+    return 0
+
+def addSavingThrowBonus(attachee, args, evt_obj):
+    if attachee.stat_level_get(stat_level_sorcerer) > 0:
+        heritage = args.get_arg(1)
+        heritageElement = heritage_feat_utils.getDraconicHeritageElement(heritage)
+        saveDescriptor = getSaveDescriptor(heritageElement)
+        flags = evt_obj.flags
+        #This is not working properly, because as soon as the spell uses reflex_save_and_damage
```
DudeMcDude, line 50: Can you give an example?

Sagenlicht: I did a post in the co8 forum, found it easier to attach all needed files there.

Sagenlicht:
> Possible bug: I don't get evt_obj.caster_class to work, it runs into an error. EventObjSpellsPerDay "spell_level", "base_caster_level" and "bonus_list" work fine.
@DudeMcDude could you please look into this, it keeps me from finishing the bonus spells. If I am doing something wrong there I apologize, but to me it seems not to work.

tpdatasrc/tpgamefiles/scr/tpModifiers/draconic_heritage.py
DudeMcDude: Should be fixed now, please help test this :)

tpdatasrc/tpgamefiles/scr/tpModifiers/bonus_spell.py
```diff
@@ -0,0 +1,68 @@
...
+    args.set_arg(2, spellLevel)
+    return 0
+
+def applyExtraSpell(attachee, args, evt_obj):
+    classEnum = args.get_arg(1)
+    spellLevel = args.get_arg(2)
+    #I am unsure why evt_obj.caster_class is not working, can't get a print of it either
+    #print "evt_obj.caster_class: {}".format(evt_obj.caster_class)
+    #Disabeld for now due to this
+    #if not evt_obj.caster_class == classEnum:
```
DudeMcDude, line 18: I've added get_caster_class() method to the event object; I think the issue was that it was directly accessing an enum field which makes it typed in python, and this causes the issue

Sagenlicht: Thanks that fixed the issue!

Sagenlicht:
I changed the spells per day behaviour to be in line with the general arcane caster class PrC handling in ToEE, so the spells per Day will automatically applied to the highest class and spell level. This simplifies the handling considerable. Should have done this in the beginning. 

Bonus Spells per Day Dragon Disciple Feature is now completed and doesn't use any workarounds.

PR #631, opened 2021-09-29
Fix indexing error in RemoveWielderCond

dolio:
The loop was using the condition being searched for to bump the argument index, not the condition found.


dolio:
Seems to be working.

DudeMcDude:
Good find! I suspect this may have been behind the recent crashiness... need to verify this.

DudeMcDude:
Nope, doesn't look like it was.
The good news is I was able to get a crash once in the tutorial section of all places, which has a fairly well defined sequence, so hopefully I should be able to reproduce it.

PR #635, opened 2021-10-02
Fix arcane archer imbue arrow not stopping bard songs

doug1234:


TemplePlus/condition.cpp
```diff
@@ -6902,6 +6902,9 @@ int ClassAbilityCallbacks::BardicMusicOnSequence(DispatcherCallbackArgs args)
 				for (int i = 0; i < actSeq->d20ActArrayNum; i++) {
 					if (actSeq->d20ActArray[i].d20ActType == D20A_CAST_SPELL) {
 						interruptMusic = true;
+					} else if (actSeq->d20ActArray[i].d20ActType == D20A_PYTHON_ACTION) {  // For spell casting equivalents, stop music
```
DudeMcDude, line 6905: I vaguely recall there was a larger issue here... @dolio I think you mentioned this?

dolio: I'm not sure. Maybe something similar came up when we were talking about concentration?

doug1234: Not sure what the issue would be.  I tried it out and it worked correctly.  Is there anything you think I should test?

DudeMcDude: No, I was just wondering if cast spell was correctly singled out here or if it’s a broader issue. Looks like the former.

I do think we should try to get a better idea what "spell equivalent action" means, though. As we've seen recently with the touch attack flag, that sort of ambiguity can create unintended effects. And in my experience, this can happen much sooner than you would think 😉

doug1234: I could change it to something like "stops bard songs".  Would that work better?

doug1234:
Should be good to do now.

PR #637, opened 2021-10-03
Handle item conditions in remove spell

dolio:
Attempting to fix the issue where spells like Magic Weapon remove item wielder conditions in most-recently-added order, rather than actually looking for the right condition.


dolio:
Seems like it's working correctly. I tested the 4 spells removing themselves, and a couple cases of out-of-order removals.

PR #648, opened 2021-10-19
Allow beguiler to take abjurant champion.

doug1234:


doug1234:
Should be ready to go now.

PR #652, opened 2021-11-13
Marshal Class

Sagenlicht:
**Known Issues:**
* Forgot to add 	AddSpellTeleportPrepareStandard and AddSpellTeleportReconnectStandard to the auras
* spell end needs review

**Used Workarounds**
* As feats can't be marked as bonus feats only, I used args to handle new aura learning. Revisited once Class Features is availible
* Spell Caster Level is discarded by the engine and set to 0. I assume because the Marshal can't cast spells and I just use the spell engine to handle the Auras and the Adrenaline Boost. Spell Level is now manually set in the "spells"

**ToDo**
* Particle effects are not completed
* Help file needs a revisit


doug1234:
Very nice!  I think you should update py00416standard_equipment_chest.py for the marshal.  Will look over it a bit more when I get a chance.

Sagenlicht:
> Very nice! I think you should update py00416standard_equipment_chest.py for the marshal. Will look over it a bit more when I get a chance.

Ty @doug1234 
Will take a look at the standard equipment chest script, ty for pointing this out to me.

PR #653, opened 2021-11-20
Unarmed Feats

Sagenlicht:
This package contains a few new unarmed feats. With the exception of Flying Kick and Zen Archery, all other feats are interacting with Stunning Fist Charges. Due to this, I created a Stunning Fist Extender and reworked Extra Stunning to work with the new extender. I think @doug1234 originally did Extra Stunning. The idea behind the new extender is basically to work like turn undead feat interaction.

The new Stunning Fist Extender does:
* ET_OnNewDay queries for for charge-increasing feats and adds them to the number of daily charges
* Reacts to a Python Signal to reduce Charges of Stunning Fist
* Answers to a Python Query that asks for remaining charges
* Adds a tooltip for remaining charges (mouseover only)

Charge Increasing Feats:
Basically all Feats from the PHB II increase Stunning Fist charges by one and Extra Stunning by 3 (the latter can be taken multiple times)

New Feats:
* Axiomatic Strike
* Fiery Strike
* Fiery Ki Defense
* Fists of Iron
* Flying Kick
* Freezing The Lifeblod
* Ki Blast
* Zen Archery

I will add the missing feats on monday, ran out of time, but I'd like to get a feedback if the changes to the Stunning Fist/Extra Stunning are ok.


doug1234:
Good stuff!  I totally agree with what you have here.  I wish I had done stunning fist the same way I did turn undead initially.

Sagenlicht:
Sadly Ki Bolt is still missing, will add it tomorrow. On the plus side, I sneaked in Zen Archery :)

tpdatasrc/tpgamefiles/rules/feats/ki blast.txt
```diff
@@ -0,0 +1,10 @@
+name: Ki Blast
```
DudeMcDude, line 1: At long last, ToEE has kamehamehas 🤣 

Only thing that's missing are EWGFs 😁 
https://m.youtube.com/watch?v=jWcAwn5LYDo

Sagenlicht: @DudeMcDude
I'd like to have a feedback on a design choice. As the Strikes will not get fired every combat round, I decided not to use an arg as activate flag and instead worked with an effect condition. This would avoid unnecessary onBeginRound and OnDealingDamage Hooks every round or damage step. Now that I know that there is actually a limit for conditions, I am not sure what's better, Avoid unnecessarcy hooks every round and use helper conditions or avoid extra conditions. If you prefer activation flags over helper conditions, I would change the implementation.

DudeMcDude: Yiu can assume that the condition limit will be overcome.

Sagenlicht:
Added Ki Blast
This packages is now finished, barring the below listed Known Issue.

Known Issues:
Ki Blast spell doesn't end properly, though I do not know why. In OnEndProjectile spell_end is called properly, spell.target_list is empty and that point and number of projectiles reduced to 0.

Ki Blast could benefit from a code review.

tpdatasrc/tpgamefiles/rules/char_class/class012_monk.py
```diff
@@ -40,72 +40,81 @@ def GetClassHelpTopic():
...
-	return
\ No newline at end of file
+    newLvl = char_editor.stat_level_get( classEnum )
+    bonus_feats = []
+    if newLvl == 2:
+        bonus_feats = [feat_combat_reflexes, feat_deflect_arrows]
+        #I know, atm a monk has to have stunning fist
+        #but maybe improved grapple will be added as a choice at first level
+        if char_editor.has_feat(feat_stunning_fist):
+            bonus_feats.append("Fiery Fist")
```
DudeMcDude, line 107: Hmm, this is new content, does it show up if new content isn't enabled?

tpdatasrc/tpgamefiles/scr/tpModifiers/stunning_fist_extender.py
```diff
@@ -0,0 +1,39 @@
...
+
+def getStunningFistCharges(attachee, args, evt_obj):
+    chargesLeft = args.get_arg(0)
+    evt_obj.return_val = chargesLeft
+    return 0
+
+def stunningFistTooltip(attachee, args, evt_obj):
+    #Adding a tooltip, so you can see how many charges are left
+    chargesLeft = args.get_arg(0)
+    evt_obj.append("Stunning Fist uses left: {}".format(chargesLeft))
```
DudeMcDude, line 30: Can you try using a string from .mes file here? Also less verbosity

DudeMcDude: Actually, please remove this tooltip, it should only appear in the radial menu. I don't want the clutter.

tpdatasrc/tpgamefiles/rules/char_class/class012_monk.py
Sagenlicht: No it doesn't show up, the feats are all flagged non core and I've tested it with Non Core options disabled, when you level up a monk you can't see the feat and therefor you can't select it as a bonus feat as well.

tpdatasrc/tpgamefiles/scr/tpModifiers/stunning_fist_extender.py
Sagenlicht: I fully agree with you, I always set charges in the radialmenu in the form of (chargesLeft/maxCharges). But the radial is done by the game atm, I would need to replace it, to add the charges to the radial correct?

DudeMcDude: Yeah i thought so.
I think I'll add an option to replace existing hooks.
Alternatively you could rewrite the whole condition from scratch if you're up to it. I think @dolio was working on it?

Sagenlicht:
Ok, so I replaced Stunning Fist. Side effect bonus: Stunning Fist does not trigger an full attack anymore, so now you can choose to just do a single attack and then move afterwards. Ofc you still can trigger a full attack normally after activating Stunning Fist.

I am unsure what bug @dolio was referring to, so I am not sure it is fixed by the redo.

Sagenlicht:
@doug1234 fyi I switched Brutal Throw to use integer 180 instead of 0 as bonus_type so it will not stack with Zen Archery
Bonus I added a help file entry for Burtal Throw and Power Throw :)

doug1234:
Sounds good.  I agree with the change to brutal throw and thanks for adding the help entry.

PR #658, opened 2021-12-30
Vigor fixes

Sagenlicht:
In my recent playthrough I used the vigor spells from Co8 the first time. The Lesser Vigor (I didn't test other ones) spell did corrupt my save files. If I cast the spell on a party member, save the game while the spell is active, and loading the save afterwards leads to a game crash, making the save unusable. I can reproduce this behaviour.

Testing was done outside of combat, on different party members and in different locations. All tests did leave to corrupted save files.

I wanted to do the mass versions as part of the Spell Compendium anyways (the normal Vigor spells are part of the SC as well) so I redid the three single target spells as well as adding the new ones. They all use the same condition (sp-vigor_fast_healing), all spells only pass their fast healing amount to the condition.

Bonus side effects:
* You can't stack different vigor spells anymore (which you shouldn't be able to do), only the highest fast healing effect will stick.
* I added help file entries for all three existing Co8 Vigor spells as well as the two new ones
* Vigor spells are now hyperlinked to help files in the History Window
* All Vigor spells now have a new particle effect
* All Vigor spells have a Vigor buff icon now
* Buff icons shows correct duration
* The Co8 spells now have a spell long description

Cons:
I have never replaced an existing spell from Co8. So I hope I did it the correct way.
* I removed the Spell794 - Lesser Vigor.py as well as the other 2 scripts from Co8infra and did regenerate that file
* I added the spells to my spell_compendium admin files with the exception of the sound file. Though I am unsure if you can actually use Temple+ without Co8. If not, not all admin file entries would be needed

Edit: These spells require my updated spell_utils


tpdatasrc/tpgamefiles/scr/spell_utils.py
```diff
@@ -421,3 +473,154 @@ def __init__(self, name, argn):
...
+    def AddSpellDismiss(self):
+        self.AddHook(ET_OnConditionAdd, EK_NONE, spell_utils.addDismiss, ())
+        self.AddHook(ET_OnD20Signal, EK_S_Dismiss_Spells, checkRemoveSpell, ())
+    def AddSpellNoDuplicate(self):
+        self.AddHook(ET_OnConditionAddPre, EK_NONE, replaceCondition, ())
+
+### Aoe Modifier Classes ###
+def addAoeObjToSpellRegistry(attachee, args, evt_obj):
+    spellPacket = tpdp.SpellPacket(args.get_arg(0))
+    conditionName = args.get_cond_name()
```
DudeMcDude, line 506: I'd add an optional arg for particles that is by default None

DudeMcDude:
Cool stuff!

tpdatasrc/tpgamefiles/scr/tpModifiers/sp_vigor_fast_healing.py
```diff
@@ -6,21 +6,42 @@
...
     ### workaround end ###
     game.particles ('sp-Vigor', attachee)
-    attachee.heal(attachee, healDice, D20A_HEAL, 0)
-    attachee.healsubdual(attachee, healDice, D20A_HEAL, 0)
-    game.create_history_freeform("{} is healed for {} by ~{}~[{}]\n\n".format(attachee.description, healAmount, spellName, spellTag))
+    if attachee.obj_get_int(obj_f_critter_subdual_damage):
+        attachee.healsubdual(attachee, healDice, D20A_HEAL, 0)
+    elif attachee.obj_get_int(obj_f_hp_damage):
+        attachee.heal(attachee, healDice, D20A_HEAL, 0)
+    if game.combat_is_active(): #Limit History output to combat only
```
anatoliy-savchak, line 24: Why?

Sagenlicht:
Removed the onRemove heal sum up again, it triggers on too many occasions that it's not worth to add the missing ticks on a rest trigger.
@anatoliy-savchak What did you comment why on? That I limited the history output to combat only or on the sum up?

anatoliy-savchak:
Yes @Sagenlicht. As a Player I want to have all logs at least in some place. History pane should have as much as it can IMHO. But it's up to you.

Moreover being DM I always ask Players to show me heal wand dice rolls and account wand leftover. So personally I would love to see any healing rolls in my History pane.

Sagenlicht:
> Yes @Sagenlicht. As a Player I want to have all logs at least in some place. History pane should have as much as it can IMHO. But it's up to you.

I am unsure about this myself. I am a huge fan of verbosing the history myself, but Vigor ticks alot, Inside combat this is no problem, I agree, information is key here, but outside of combat it appeared a little bit spammy to me while I playtested Vigor in my current runthrough.

In addition I am not sure how confusing the new behaviour now actually is, when the history output suddenly stops...

tpdatasrc/tpgamefiles/scr/tpModifiers/sp_vigor_fast_healing.py
```diff
@@ -6,21 +6,42 @@
...
+        game.create_history_freeform("{} is healed for {} by ~{}~[{}]\n\n".format(attachee.description, healAmount, spellName, spellTag))
+    return 0
+
+#Adding an onRemoveHealTick check for triggering a Rest/Passing Time
+#While a vigor effect is ticking; This will sum up leftover ticks in one heal
```
DudeMcDude: IIRC the onBeginRound event also tells you how much time has passed so you shouldn't need this workaround

Sagenlicht: > IIRC the onBeginRound event also tells you how much time has passed so you shouldn't need this workaround

Here is what I tried:

- My first idea was to simply to add a hook for ET_OnNewDay. To me, this seemed to be the hook use, as it avoids any other reason the spell condition is removed (like getting killed or the effect getting dispelled) but it didn't work as the spell condition gets removed before the hook actually gets triggered by the New Day Event.
- Then I tried to modifiy the already existing ET_OnBeginRound hook by simply adding an extra check (if evt_obj.data1 != 1) but to my surprise the OnBeginRound Hook is not triggered when you hit the rest button.
- Then I tried to use the ET_OnConditionRemove Hook, but this adds way to much headache, as it is triggered by too many events (e.g. TeleportReconnect, Dispel, etc) so I would need alot of case handling just to handle an edge case. This didn't seem worth it.

So atm I did scrap the heal sum idea, I admit I still would like to have it, but I am out of ideas how to do it properly.

PR #663, opened 2022-01-27
Fix to allow elves to be abjurant champions

doug1234:


doug1234:
This is pretty safe and its been available for review a while so I am going ahead and merging it in.

PR #664, opened 2022-02-12
Update to_hit_processing.py

Sagenlicht:
Fixed logbook entry for misses
Fixed crit confirmation rolls


DudeMcDude:
Thanks! Funny how that went unnoticed for so long...

PR #669, opened 2022-03-29
Ultimate magus Prc

doug1234:
Still needs documentation but should be ready otherwise.


DudeMcDude:
As of next version, I want to split off features from fixes. 

So before merging to master, please go through the features branch.

On that note, I think it's time to finally have it pack tpdatasrc while building as you had previously suggested... the current situation is a problem.

doug1234:
Sounds reasonable to me.  How do you suggest we add the ability to pack while building.  Do you think its OK to add datapack to the repo or do you have something else in mind?

DudeMcDude:
> Sounds reasonable to me. How do you suggest we add the ability to pack while building. Do you think its OK to add datapack to the repo or do you have something else in mind?

Yeah go ahead.

doug1234:
It does have both dlls in the same directory as the executable.  Here is what I get when I run:

D:\Games\TOEE\Modding\TemplePlus-Clean\tpdatasrc>D:\Games\TOEE\Modding\TioTool\TioPack.exe tpgamefiles.dat tpgamefiles
[INFO] Creating database file tpgamefiles.dat...
[INFO] Using compression level 9...
[INFO] Building file list...
..Miserable Death...

PR #670, opened 2022-04-06
Added back data pack.

doug1234:
Checking to see if it causes an issue with CI.


doug1234:
Now builds correctly in CI and through running Visual Studio.  Should be good to do.

DudeMcDude:
Have you checked that the files are valid? IIRC currently the infrastructure uses the recreated dat packer, which I think still had some diffs with the original one.

doug1234:
I will take a little more them then and double check that the dat files made by the CI work the same.

doug1234:
I checked that the ones in the artifact loaded.  It turns out I have been using the packer from dependencies for a while now and haven't had any issues.  Is there anything in particular you think I should test?

DudeMcDude:
I just vaguely remember doing a bitwise comparison between our datpacker and the one that relies on tio.dll and seeing some differences. It could be that:
1. It got fixed in the meantime (by @shartte), so there are no more differences.
2. Errors are still present, but are subtle.
3. Differences exist, but they are meaningless.

So I guess the first step would be to check if there's still any diffs :)

DudeMcDude:
Locally you mean?
Just to state the obvious, does it have tio.dll and the zip dll in the same folder?

DudeMcDude:
Ah, one thing I can think of - which version of tio.dll are you using? It needs the one dated 2004-10-12 (CRC2 = 2E1CB46B, available in GoG, not the patch2 version)

doug1234:
The dll is indeed dated 2005-10-12.

DudeMcDude:
Alright i finally got around to comparing the results myself. Looks ok (as far as i can tell the diffs are just due to uninitialized data in fixed size null terminated strings, which should be of no consequence).

doug1234:
Thanks a ton for doing that!  This should make things easier going forward.

PR #674, opened 2022-04-14
Fixed bug where wizard oposition schools effected other classes

doug1234:


DudeMcDude:
Btw in sagenlichts branch we've fixed an issue when spell school was null, can you merge that in too?

doug1234:
I think I added the change you were talking about.  BTW:  How should we move toward getting more stuff into the feature branch.

Issue #680, opened 2022-07-08
-4 on attack with touch and ranged touch


zicion:
Im playing with both temple+ and Co8 expansion and when i attack with spells im getting a -4 Option: Dealing Nonlethal Damage

![unknown](https://user-images.githubusercontent.com/74786857/177993857-e3c23805-1890-44dc-bd7c-0131518883c1.png)
g

doug1234:
I couldn't reproduce this immediately and in the code it looks like the penalty should only be applied when a weapon is used. Please give some more details.  Was the non lethal damage option selected?  What spells did you use?  What weapon was equipped?

DudeMcDude:
Is the "deal normal damage" option ticked in the radial menu? What spell is this?

zicion:
![image](https://user-images.githubusercontent.com/74786857/178104983-b22a739c-5796-495c-b34f-f55c1e3b3499.png)


This is my stat block in case you need it, the weapon i have equipped is a rapier but it doesn't matter if im doing it with another weapon or barehanded. Spells both ranged and touch always have this -4, and i dont have non lethal damage on. Any way to share the save file? I can post it on the Circle of 8 discord just PM me zecond#0428

zicion:
Just found out why its doing it, when i have "cast defensively turned" on, there is no mention of this -4 malus in the Help menu.

doug1234:
Cast defensively doesn't seem to make a difference for me.  I wonder if this has something to do with the ranged sneak attack option since you have Arcane Trickster levels.  Please send a save file to:  templeplushelp@gmail.com

doug1234:
Never mind... I just reproduced it.  Unarmed and deal normal damage on were the key.  Arcane trickster levels were not necessary and cast defensive didn't seem to make a difference for me.

doug1234:
Looks like a bug in vanilla TOEE in function 100F9040.  I don't see anything checking if it was a ranged attack or touch attack.  I should be able to fix if pretty soon.

PR #681, opened 2022-07-09
Fix issue #680.  No deal normal damage penalty for spells.

doug1234:


doug1234:
This is also a bugfix.

PR #719, opened 2022-10-21
war domain feats

Firemephit:
Added weapon proficiency/focus feats for creation/first level of cleric (it's a little messy with a massive switch statement but it works.)


DudeMcDude:
Check out feats.GetFeatForWeaponType

PR #728, opened 2023-04-18
Some animation related changes

dolio:
Looking into some animation related stuff. The first is an attempt to see if it's possible to match weapon animations to the logic that determines whether they're considered to be wielded 2-handed (for e.g. damage purposes). This involves:

- Moving the logic for determining whether a weapon is being wielded two-handed into the inventory module, and just referring to it in the original condition location.
- Modifying the animation lookup to use the same one vs. two handed logic. The animation arrays do seem to have both 1 and 2 handed references for many weapons.

Hopefully this will just cause the game to e.g. play the 2-handed animation if you're wielding a long sword without a shield/offhand weapon.

### DIFF
--- a/TemplePlus/animgoals/anim.cpp
+++ b/TemplePlus/animgoals/anim.cpp
@@ -409,28 +409,6 @@ bool AnimSystem::PushWalkToTile(objHndl handle, LocAndOffsets loc){
     return false;
 }
 
-bool AnimSystem::PushKnockback(objHndl handle, LocAndOffsets loc) {
-	if (!handle || critterSys.IsDeadOrUnconscious(handle))
-		return false;
-
-	if (!critterSys.IsPC(handle)) {
-		auto getDlgTarget = temple::GetRef<objHndl(__cdecl)(objHndl)>(0x10053CA0);
-		if (getDlgTarget(handle) != objHndl::null)
-			return false;
-	}
-
-	AnimSlotGoalStackEntry goalData(handle, ag_attempt_move_straight_knockback);
-	AnimSlot *knockSlot;
-	goalData.targetTile.location = loc;
-	if (!Interrupt(handle, AnimGoalPriority::AGP_5, false))
-		return true;
-
-	goalData.Push(&animIdGlobal);
-	GetSlot(addresses.animIdGlobal, &knockSlot);
-	knockSlot->path.flags &= ~PF_COMPLETE;
-	return true;
-}
-
 void AnimSystem::SetRunningState(bool state){
     AnimSlotId *runId = addresses.animIdGlobal;
     AnimSlot *runInfo;
--- a/TemplePlus/animgoals/anim.cpp
+++ b/TemplePlus/animgoals/anim.cpp
@@ -419,12 +419,12 @@ bool AnimSystem::PushKnockback(objHndl handle, LocAndOffsets loc) {
 			return false;
 	}
 
-	AnimSlotGoalStackEntry goalData(handle, ag_knockback);
-	goalData.targetTile.location = loc;
-	if (!Interrupt(handle, AnimGoalPriority::AGP_3, false))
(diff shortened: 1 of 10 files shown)


dolio:
It works. Character is wielding a long sword.

[two-handed.webm](https://user-images.githubusercontent.com/750648/233243389-2f2ab61f-89ba-4faa-aa28-965729fb300f.webm)

dolio:
Here's a couple things I could use some input on if anyone's watching.

1. `anim_goal_push` works, but certain animation ids crash the game very hard. It seems to completely bypass the crash dump somehow. I can guard against some of these, but maybe something else should be done (as well)?
2. I'm not sure if anyone else has ever noticed, but if you take Improved Unarmed Strike, you start acting like a monk, even if you have no monk levels. I'm kind of tempted to switch the check to testing for actual monk levels, so you could in principle have an unarmed brawler who isn't doing axe kicks and stuff. Any objections to that?

anatoliy-savchak:
> Here's a couple things I could use some input on if anyone's watching.
> 
> 1. `anim_goal_push` works, but certain animation ids crash the game very hard. It seems to completely bypass the crash dump somehow. I can guard against some of these, but maybe something else should be done (as well)?

IMHO you should create AnimGoalPushSafe if you are to expose this func. Suppress unsafe ones.

dolio:
Here's something else I'd like to fix, but I'm unsure how. If you fool around with `anim_goal_push`, you'll find that there are some cool animations for conjuring spells (this one is enchantment, I think):

[conjure.webm](https://user-images.githubusercontent.com/750648/233863013-4daf62ca-0cea-4616-ba90-2edfc84e4ef0.webm)

I got that loop by manually timing calls to `anim_goal_push` in the console. These animations are, I think, actually used when casting spells. But it seems like the timing is off. When you cast a spell normally, it starts that animation, but then almost immediately goes into the 'finish casting' animation. And when you're _interrupted_ during casting (by a readied action), it seems like the game pushes this animation repeatedly like 5+ times/second. So instead of getting that nice animation loop, the caster just twitches weirdly.

I've been reading animation code, but it's not clear why this happens. I'm not even sure why the game sometimes chooses to play animations in a loop, and sometimes doesn't. Does anyone know more, and where this might be fixable?

DudeMcDude:
Oh hey, didn't notice this until now. Nice work!

Are you still debugging via appveyor builds?

dolio:
Yeah, I never figured out how to build it locally. :stuck_out_tongue: 

Although it's now been a while since I tried, so I don't know if anyone's done anything to make success more likely.

DudeMcDude:
> Yeah, I never figured out how to build it locally. 😛
> 
> Although it's now been a while since I tried, so I don't know if anyone's done anything to make success more likely.

Oh, nothing has changed with that.

Sorry that i can't help more, my main dev machine died... also still addicted to chess :D

dolio:
I can't rightly complain. I disappeared for like a year, distracted by math stuff. :)

doug1234:
I was wondering i this is complete or WIP currently?

dolio:
Think it's good to go. I had one change hanging out in my local copy: using sword animations for kukri attacks, since they're slashing. But that's pushed now.

I think at some point I went through all the weapon animations and checked that they made sense, but it's easy to tweak now if anyone notices cases I overlooked.

doug1234:
Thanks!!  Do you have a set of release note bullet points for the animation fixes?

dolio:
Sure.

- Fixed animation logic to make 2-handed animations correlate with game mechanical 2-handed attacks (like, wielding a longsword 2-handed)
- Audited weapon animations and swapped to more sensible choices for some weapons (like, use a thrusting animation for rapiers)
- Added an `anim_goal_push` python function to start arbitrary animations (by id)
- Added a `turn_toward_loc` python function for orienting a character toward a location rather than another object.

PR #749, opened 2024-01-11
Fix some weapon animation bugs caused by the 2-handed patch

dolio:
This PR fixes two bugs introduced by the previous weapon animation patch.

1. Brother Smyth wasn't properly doing his forge animation, because he's wielding a war hammer. The fix to this is to prefer 1-handing/the old behavior for `WeaponAnim::SpecialN`.
2. Wearing a buckler while 2-handing weapons was giving weird results. It looked like you were just punching people while holding the weapon. Here the problem seems to be that packing an off-hand animation (for the buckler) doesn't work for 2-handers. You need to have the 2-handed animation packed in both spots. So I skip the off-hand logic completely if the main hand is being 2-handed.

I also did a little work to figure out what a `3` result from `GetWieldType` is supposed to mean, and put it in a comment specifying its results.


doug1234:
Good find.  Maybe I need to put out a small bug fix version soon.

PR #752, opened 2024-01-20
Topic/double weapons

### LINKED ISSUE #750: AC enhancement bonuses aren't implemented correctly
While messing around a few days back, I realized that various AC enhancement bonuses aren't actually implemented properly. Here's the way it works right now:

- All AC bonuses are put in a bonus list with their respective types
- Base armor bonuses from worn armor have a different type than enhancement bonuses to that armor, which allows them to stack
- All the bonuses are added up using the usual 3.5 rules of bonuses of the same type failing to stack

This might seem like it would give the correct result, but it actually doesn't. You can observe the results this way:

- Wear a chain shirt +3, which should give 7 AC
- Wear bracers of defense +6 which give 6 AC

Tabletop rules say this should give  you +7 AC from armor. The enhancement modifies the chain's base 4 AC to 7, which is greater than the 6 from the bracers. However, in ToEE you get +9 AC, because you get the +6 base armor bonus from the bracers (which is grater than the +4 from the shirt) and the +3 enhancement bonus from the shirt still applies.

I suspect this also happens with shields. So if you wear a +3 buckler, casting shield probably gives you +7 shield AC instead of +4. It would also happen for natural armor, but I don't think there are ways to get multiple sources of natural armor; all the spells and items give enhancements, and the base just comes from the object fields.

I'm not sure how to fix this best. I'll think it over.

dolio:
This work is getting pretty close, so I figured I'd start writing it up.

This adds features and fixes related to weapon wielding. The major feature is enhanced options related to two weapon fighting.

1. There is a toggle radial for controlling whether you actually use two weapon fighting. This lets you do the following:
    - Turn off two weapon fighting even while holding two weapons, to avoid taking the two weapon penalties
    - Turn _on_ two weapon fighting when wielding an appropriate shield to use shield bashing as a second weapon
    - Turn _on_ two weapon fighting while holding a double weapon
2. There is a toggle radial for swapping which hand is your primary weapon. This lets you do the following:
    - Attack with just your left hand if two weapon fighting is off, so you can fight with just your left hand without a penalty
    - Shield bash as your primary attack
    - Use the opposite end of your double weapon as a primary attack (only affects animation at the moment)

The toggles automatically reset themselves when you equip weapons to get the old behavior as a default, so people should only need to interact with them if they want to be fancy.

Shield bashing causes you to lose the AC bonus from the shield until your next turn unless you have the Improved Shield Bash feat, but at the moment there's no way to get that feat (just a check for it). Shield bashing requires martial weapon proficiency, but I didn't add a specific proficiency for it, because I suspect no one would take it. I just check for the ALL proficiency.

As part of this I moved around and deduplicated a fair amount of weapon wielding logic. I also replaced some original toee functions after reimplementing/fixing their logic.

While I was digging around in stuff, I also made the following tweaks:

1. Implemented some additional strict rules cases for weapon proficiencies
    - Dwarves were getting waraxe proficiency for free, when they're only supposed to treat 1-handed wielding as martial
    - Gnomes were getting hooked hammer proficiency for free, when they're supposed to treat it as a martial weapon
    - Half-orcs were getting orc double axe proficiency, but the books don't say that. Orcs weren't getting to treat them as martial weapons.
2. Fixed the way enhancement bonuses to armor and shields work, to prevent stacking and erroneous information in the UI. This fixes #750 
3. Select some weapon animations based on how you're wielding them at the moment.
    - Double weapons use the staff animation _while you're two weapon fighting_.
    - When you're two handing them, double weapons use hammer or axe animations. This seems kind of accurate for staves, based on some brief youtube searching.
    - Maybe I'll also do the proper toggle for selecting between two damage types for e.g. daggers. Then use different animations based on how you're trying to damage the enemy.

I think that might be all of it. I still have some finishing touches to put on it, but I'll also post some videos showing off what you can do.

Also, when this is merged, it might be better to squash it. My CI-based development methodology causes a lot of meaningless commits.


dolio:
Two weapon fighting with a quarterstaff (note: the AoO doesn't happen anymore):

[quarter.webm](https://github.com/GrognardsFromHell/TemplePlus/assets/750648/62d99b88-a23e-401f-9feb-476eb0b20568)


Two weapon fighting with a shield bash:

[bash.webm](https://github.com/GrognardsFromHell/TemplePlus/assets/750648/ef5cdce2-a5cf-4b88-ac8e-da75e22456fa)

Shield bashing as a single attack:

[bash-only.mp4](https://github.com/GrognardsFromHell/TemplePlus/assets/750648/8db971dd-5dc7-4a9e-8f30-0e55a55a99bc)

dolio:
Here's the two handed quarterstaff animation:
[two-quarter.mp4](https://github.com/GrognardsFromHell/TemplePlus/assets/750648/5b52d3f3-53d0-4187-8516-cd2b4d0b5eae)

If anyone can think of which animation might look better as actual quarterstaff fighting, let me know. I think polearm could actually look better, but I think the staff would be held in the middle, which would look silly (I think the place you hold it is a function of the quarterstaff model, not the animation).

doug1234:
Wow this is really impressive!  I will look the changes over in detail when I get a chance.  Are the armor/shield and enhancement bonuses broken out when they are displayed?  Or are they combined as an armor bonus on the display?

I can add the Improved Shield Bash feat if you don't get around to doing it.

dolio:
They're combined. So, if you have +3 full plate, you'll see a +11 armor bonus to AC in the list, which overlaps with +5 from bracers, say. That's basically how all 'enhancement' bonuses to AC work, in a way that's, I think, unique to AC. You have various bonus types to AC (armor, shield, natural, ...) and you can get enhancements _to those bonuses_, which modify the number before adding it to the AC bonus list.

It now matches what you see when hovering over the armor in the inventory, too. Full plate +3 was always just saying 11 AC there. I suppose I could actually give a more detailed breakdown there, because the information is now available where the info is formatted. But I don't know if anyone would care about that breakdown.

dolio:
Oh, by the way. One of the things I haven't done in this PR yet is actually make shields able to be bashed with.

The way I set it up, you need to put an item condition on a shield to allow it to be used to bash. I did this so that it would be obvious which shields _can't_, like tower shields and bucklers, and also because the condition needs to supply some of the weapon stats, because I think armor doesn't actually have memory allocated for all the weapon stats.

This means that I need to change proto lines for all the relevant shields. Does that go in `co8fixes` and `kotbfixes`, in the `protos_override.tab` file?

doug1234:
Ah... let me know how it does letting the shields be bashed with.  I am pretty sure that protos_override.tab in co8fixes and kotbfixes is indeed what you want.

dolio:
Right now it requires replacing proto lines for the relevant shields with lines that are identical, except they have an extra condition for shield bashing.

Oh, also, I think I added some help entries in one commit, but I think they're not showing up in game. I added them in `tpdata/tpmes/help_extensions.tab`. But I noticed there are also `help_extensions.tab` files in the `co8infra` and `kotbfixes` directories. So, like, are only the latter relevant or something?

doug1234:
You want them in co8 and kotb.  I assume shield bash is something we would want for all version of the game.  Maybe they should be in tpdata also which I think applies to the unmodded version of the game.

dolio:
I decided to hook into the proto parser so that you can write down the relevant weapon stats for a shield, and it will generate the corresponding shield bash condition on the item. I also tweaked the shield bash condition for this, so it uses the designated size for determining whether it's a light weapon, and so that it supports all the relevant weapon fields. This means I can make the spiked shield do piercing damage like it's supposed to.

I also added stats for the Co8 protos_override.tab. It seems like shields are systematically one size too large for some reason, so I had to revise that part, too. I didn't bother with the gnoll shields because it seems like those aren't meant for PC use.

doug1234:
Cool stuff.  Is it possible to target the spiked shield with spells that effect weapons like magic weapon?

doug1234:
I was also wondering if the quarter staff defaults to two weapon wield?

dolio:
All double weapons default to their normal behavior.

You can target a shield with magic weapon, but it'll just fizzle. I think each spell needs to be updated to work, because the logic for that is in the spell scripts. If you successfully get the effect on the shield I think it'll just work, because I didn't need to do anything special to make the holy/unholy shield to work.

doug1234:
"On a tangent, in theory you're supposed to be able to separately enchant each end of a double weapon, but right now it doesn't work that way. I think you're actually obliged to enchant both ends separately, but that seems like a dumb rule, because being half the cost of two weapons is the only mechanical advantage I can think of for double weapons."  I am not quite clear on wht you are saying.  Are you saying TOEE isn't working like the SRD says it should?  Is it currently forcing you to enchant both ends the same way?

PR #758, opened 2024-02-12
Various minor fixes

dolio:
Here's a collection of minor fixes and tweaks.

1. Dismissing a spell is a standard action under strict rules
2. Makes the `spell_radius` member of a spell object account for widen spell. This doesn't matter often, because the picker has already captured the right targets for instantaneous effects, but persistent area effects need to create the right size object event.
3. Adds an additional optional argument to `conditions_get` that lets you filter to (in)active conditions. Default is all conditions.
4. The `Tactical` radial menu enum in Python was mistakenly named `Tacitical`. As far as I can tell it was never mentioned anywhere, so I just fixed it.
5. Adds an animation for dismissing a spell. Might tweak this a bit more later.
6. Skill point penalties were off at level 1. Odd numbers were rounding toward zero instead of down.
7. Disarm, trip and unarmed strikes are only supposed to provoke AoOs from the target, not everyone who threatens
8. Made finger of death damage negative energy type. Unspecified seems to get cancelled by _all_ damage reduction.
9. Remove +4 ranged attack bonus against helpless targets. (Atari bug 30)
10. Fixed failing to copy one scroll letting you retry others. (Atari bug 124)
11. Treat sleeping/held (helpless) critters as having 0 dex, held as 0 str. Unconscious already did this, but it was missing from these others. (Atari bug 31)
12. Require 100GP/spell level for copying scrolls under strict rules.
13. Tweaked the copy scroll action check function to enable a better float message if you are being blocked by a previous failure-to-copy. The logic was already there, but you'd never hit it because of the way action checks work (you'd just see a generic 'Invalid Action' float).
14. Make coup de grace actually give a saving throw. Previously it was just automatically killing the target as long as you did more than 0 damage.
15. Make troll (and other?) rend attacks 2d6 + 1½ str instead of fixed 2d6+9. +9 is just the pre-calculated value for the typical troll stat block.
16. Hide 'Held' effect indicator when character also has freedom of movement. The other tooltip already did this.
17. Don't do begin-round break free from hold effects when character has freedom of movement; if you succeed, it wastes your turn. Atari bug 119.
18. Remove +4 bonus to avoid being tripped from Improved Trip. Atari bug 34
19. Enable sneak attack on Color Spray stun. Atari bug 42. (Stunning fist seems okay already) Also made the blind effect remove your action, because you're supposed to be stunned as well (there are up to three individual conditions applied in sequence for the effects, rather than up to three simultaneous conditions).

I'll write up in-game help text for some of these. I already wrote some for the spell copying stuff, but haven't committed it yet.


doug1234:
Wow... this is some really cool stuff.  Let me know when you are done tweaking.

PR #762, opened 2024-02-20
Fix crash from shield bash and mirror image

dolio:
This fixes a crash I noticed when you try to attack a mirror image while holding a shield with the new shield bash stuff. The mirror image logic ran a to hit query with no attacker/defender set, and the shield bash stuff was expecting that the attacker would be there.

I double fixed it, just to be thorough. The shield bash query checks to make sure the attacker is actually set, and the mirror image code now sets the attacker.

Hopefully I submitted this against the right branch. I noticed you'd been switching them to this one. Let me know if I need to redo against master.


doug1234:
OK cool... you are correct, I am getting stuff ready for the next release in the features branch.

PR #765, opened 2024-02-24
Beef up the hydra

dolio:
This PR enhances the hydra's abilities. It won't affect the single hydra hand placed in the temple (that apparently is in a .mob file, and has a bunch of its stats overridden from the proto), but I think it might affect a random encounter containing hydras (I've seen it in the files, but I don't remember where it occurs). The changes are as follows:

- Fixed fast healing. Some other enemies already had this (like quasits), but it was unlikely to do anything.
- Enhanced the full attack cost logic to allow taking full attacks as a standard action based on a query. The hydra has a condition that allows this, so it can move and full attack. Seemed like the easiest way to accomplish this aspect.
- Enhanced charge attacks in a similar way, so that based on a query, a charge puts you in the full attack state at the end. Should be useful for the Pounce ability as well. I think the AI uses this correctly, but I'll double check.
- Tweaked the combat reflexes logic to give hydras an extra AoO for each one of its (current) heads, instead of based on dexterity. (Atari bug 275)
- Made a python condition for hydras that hooks into the above.
- Added a proto line override in the co8 directory. Gives the hydra fast healing 15 instead of regeneration 5 (or poison, from the original game), and gives them the hydra condition with 5 heads. I also gave them the boar AI so that they charge at people.

I think the 5 headed hydra is still not very dangerous, even with all these fixes. I've been testing with 9 and 12 headed hydras, and those are pretty dangerous; they basically take out one character per round if they can reach them, and they're harder to tie up. Unfortunately we don't have models for those. (If someone could put that together, with ways of dynamically varying the number of heads, that'd be pretty sweet for the future.)


dolio:
I'm getting a link error against the python library on my fork. Do I need to update something?

doug1234:
That is odd... I wouldn't expect you to get any link errors.  What exactly are you seeing?

doug1234:
That actually does look like the typical "CI upgraded something" error.

dolio:
Well, in the mean time, I was wrong about the AI actually charging with a full attack (it seems like it never makes use of left-over attacks). So I'll see if I can do something about that. Otherwise I'll get rid of the boar AI, because charging is worse than the hydra's normal move-and-attack.

doug1234:
OK cool.  It may be tomorrow before I get my VS updated and I can see if I need to rebuild the dependencies package.

dolio:
Never mind. I was running the wrong build. The AI _does_ charge+full attack, so no fix is necessary. (The other stuff I was seeing is a separate issue, I guess.)

Take your time, though.

anatoliy-savchak:
I'd like to add certain things to the table:

1. There is extraordinary monster ability Pounce which allows full attack after move. Which brings to second point.

>  MM1 pg 313, **Pounce (Ex)**: When a creature with this special attack makes a
charge, it can follow with a full attack—including rake attacks if
the creature also has the rake ability.

2. You need to know D&D 3.5 rules. This is one of the reasons why I personally became Dungeon Master - to understand and practice rules.
3. The Hydra does not have Pounce, therefore this fix should not be here.
4. But I support creating possibility for Pounce. As there are multiple ways to achieve this marvelous ability: Claws of the Leopard from CAd, or CCh Spiritual Totem Lion Totem.
5. Hydra's Combat Reflexes is indeed problematic. It's either Head Count * Dex Mod and once per aoo, or during single aoo strike Head Count times. I would settle for first for now.
6. You need to in-depth know how tactics works in ToEE.

anatoliy-savchak:
Huh, indeed, thanks!

Then it is sort of Pounce. In that case we should implement Pounce.
It's a pity that there is no room in D20CAF.

dolio:
The rules for the hydra's combat reflexes are not super spelled out, so some interpretation is necessary. I think the people who think it's supposed to mean that a single AoO uses all the heads at once (which you often see on internet forums) are not thinking it through.

One of the big mechanics for hydras is chopping their heads off with sunder attempts. Unless you have improved sunder, each one of those provokes an AoO from the hydra. However, if it only had 1 AoO per round, this would make beheading it pretty safe. Even if it got 2 with combat reflexes (because it only has 12 dex), you could still get multiple free sunder attempts at the higher end of things. So, instead it gets a number of (otherwise normal) AoOs proportional to its current heads, so sever attempts are unlikely to ever be free (without improved sunder).

At higher levels, a full suite of attacks from a hydra can completely kill even a robust character, so the other version would turn into sacrificing a character to get free attempts from others, or just being completely unable to make attempts without improved sunder. I don't think that's the point of the ability.

I'm not in inclined to multiply the heads by the dex bonus, because there's no indication that it's supposed to work that way. I think it's good if it's just using the head count in place of the dex bonus, under the premise that each head is acting sort of independently (same as for the other abilities), and the combat reflexes feat is just to help motivate and explain the ability. That's why I made it a number of _extra_ AoOs equal to the head count, too. One for each head doing its own thing, plus the normal one.

Pounce can be added later. It's easy to do with the way I set this up (just make a condition that responds to the right signal). The instances I can recall enable extra attacks beyond the normal full attack, which requires extra work. And it's not work that has anything to do with hydras.

doug1234:
CI should be good now thanks to help from @shartte !

dolio:
Okay. Branch builds, and seems to work without my dropping the files into an override.

doug1234:
Looking good now... any way to override the mob file and have this work for the the single hydra in the temple?  Also, is the fast healing fix going to work correctly with all feat healing abilities?

dolio:
I have no idea how mob files work. They're binary, so you need one of the modding tools to edit them. I don't know if Temple+ is set up to pull them from override directories or not. Apparently even the Co8 mod isn't overriding it, because it still has poison damage instead of the `Regeneration 5` condition in the protos.

I'm not sure what feats you mean. Fast healing is supposed to be like natural healing, except you heal every round instead of every day (or hour or whatever for nonlethal). I don't know what feats it would interact with.

anatoliy-savchak:
Option A: add first heartbeat to proto and setup conditions dynamically.

Option B: use Ed to edit Hydra mob on the map. But saving the map could introduce a lot of changes. So backup the mob somewhere, restore whole folder and overwrite hydra mob file only.

dolio:
I don't think A will work. I tried changing one of the scripts in the proto, and it didn't change on the placed creature. So, I think the mob file is overriding all the proto scripts. Probably not surprising, because the placed hydra is not hostile on sight (overriding the KOS script), and I bet you have to either override none or all of those scripts.

Does the editor tell you which mob file is which placed object? The file names are like meaningless UUIDs. The only clue is which map it's in.

Anyhow, this isn't something I can easily fix. If someone else wants to pick that up, that'd be cool. The relevant differences I know of are:

- `Monster Melee Poison` with argument `4` should instead be `Monster Fast Healing` with argument `15`
- It should have an additional condition: `Monster Hydra` with argument `5`

The simplest way would be to make the mob file pull the conditions from the proto, but I don't know if that's possible without messing something else up.

anatoliy-savchak:
My first pet project for toee was creation of mob parser in c#. I have cmd which converts mob to xml and back.

I will take care of the mob.

In my modules I never use monster mobs. There is only one on the map, which create all others in Py, based on Sitra's daemon idea.

doug1234:
I said feats but I guess it is more spells that I was thinking about.  Sorry for the confusion.  Sounds like your fast healing change won't cause any issues.

@anatoliy-savchak Can you submit another PR with the mob file fix?  That probably isn't appropriate for this PR.  I am going to go ahead and merge it.  Thank you @dolio !

PR #773, opened 2024-03-30
Various changes related to spells that modify a critter's size

dolio:
This started with an aim to prevent Enlarge Person and Righteous Might from stacking, but I ended up doing a lot of work to make the overall system (hopefully) work better. Here's the overall changes.

- Added a couple new `BonusList` functions
   1. gets the largest penalty, which I ended up not needing, but someone might
   2. a special function used for model scale queries, which treats the 'initial value' differently from the rest of the bonuses. I don't think you ever _see_ the bonus list for these queries, so hopefully it won't lead to confusion
- Replaced some more standard functions that test conditions by pointer. Doing `ExtendExisting` can break these unless you try a `GetByName` using the name from the old struct. But not all the functions that include these tests were fixed.
- Tweaked the size category callbacks for spells that enlarge/reduce characters. They now use `data1` and `data2` to keep track of whether the size has already been enlarged/reduced once. This fixes the stacking of enlarge with righteous might w/r/t size category
- Replaced the on-add/remove behavior for enlarge/reduce/etc. to just trigger a model update. They no longer set object fields; all the changes are determined dynamically.
- Likewise, removed the Co8, player-only reach modification for enlarge. It doesn't get set, and removes itself on the first BeginRound if you load a save game.
- Added `dispTypeModelScale` callbacks to the relevant conditions
    - this uses the second `BonusList` function above. The initial value is set to the object's normal model scale. Then, the conditions add `+1` or `-1` bonuses. At the end, the extra bonuses are summed, and used as an exponent to scale the initial value. So, this ends up as `init*1.8^(overall bonus)`. This fixes Enlarge Person and Righteous Might stacking w/r/t/ model scale, because the bonuses overlap.
- Changed the weapon dice queries for these conditions. The query already had a bonus list in it, so the conditions just add a +1 bonus or -1 penalty. Then the _query_ uses the overall bonus/penalty to figure out how to adjust the dice. This fixes the stacking for this case (bonuses overlap again).
- Changed the `GetReach` function to dynamically figure out an adjustment to the critter's reach by using size category queries (both base and current).
    - Critters that become tiny now get decreased reach. The area they threaten is comparable to a single 'square', which I think is P&P accurate. This is from the perspective of being able to AoO enemies that enter your square, for instance.
- Changed the `GetRadius` function similarly. This was not being handled at all before, so being enlarged didn't really make you occupy more space, strangely.
    - While testing this, I've come to the conclusion that _many_ radii in the protos are just wrong. The values often don't even make sense from a visual perspective. So there are some clamps in the code to prevent really stupid numbers. Like, if you reduce a dire wolf, which has radius 30 (inches) despite being large, you don't want to get a radius of 5.
- Uniformized the stat bonus types for enlarge/reduce/righteous might stuff, so that they don't stack
- Removed the old model scale fixes, because they're no longer needed.
- Systematically replaced one of the `stat_base/level_get` cases for stats with the same category as size and stuff. They should now properly report the base and current value respectively, which was dicey before. I also renamed the category to "Physical," which is suggested by the original dll, and seems better than "Type3" or "Combat" (since not all the stats are really combat related).
- Tweaked Reduce Animal spell stuff:
    1. It's supposed to be 1hr/lvl, not 1min/lvl
    2. It's supposed to work only with willing animals

I think that's about it. This _should_ fix atari bug 138 (stacking enlarge w/ righteous might).

A couple things I haven't done, that I know about.

1. If you load a save with the Co8 stuff applied, you'll probably have the wrong reach until your next turn. I can add extra checks to try to work around that, but I'm not sure if it's a big enough deal to justify it.
2. Reduce Animal enters combat even though you can only cast it on a willing target. I forget if that's hard coded somewhere, but changing the AI designation in the spell info didn't fix it, at least.


doug1234:
Nice change!  I will take a closer look when I get a chance.

doug1234:
Saw that you other changes was done.  Is this one also ready to be merged into the features branch for a beta?

doug1234:
Please take a look when you have time. I will try for a beta in late January.  I am out of town right now too.

dolio:
Okay. I took a look, and updated this to merge cleanly.

I think it should be safe to include. Before I was worried about some of the stat stuff I'd changed, but I think most of it doesn't really affect gameplay. It'd probably be good to tell people to mess around with enlarge/reduce when you put out the beta testing call, though, just to see if they notice any weird stuff with combat behavior (like, clicking on someone putting you at the right distance to actually attack). I think I worked out the kinks of that, but it's possible I missed something.

----

The main thing I see that I failed to accomplish is people reduced to 'tiny' size taking opportunity attacks when approaching someone with a normal reach. In tabletop with a grid, this is explained by the tiny character having to enter the opponent's square. In ToEE, the movement is fine grained, and it's determined by character having to move through the other character's reach radius. So, like, when a medium character attacks someone with 10 ft. reach, they naturally move through threatened space and take an AoE. But it doesn't happen with tiny+5ft. reach.

I think maybe the movement pathing would have to have increased resolution to make it work. Like, it only checks for AoEs on a ~5ft basis or something. But I didn't figure out how to do that (and you'd probably want to only do it for small radius critters). Anyhow, it's not a regression, just something for future work.

doug1234:
Can you give a summary of these changes for the change log.

PR #774, opened 2024-03-31
Magic Missiles + Mirror Image

dolio:
Here's another quick one.

With strict rules enabled, it makes Mirror Image interact with Magic Missile. Essentially, all the images are valid targets, so the caster of magic missile has to randomly determine their targets just like other attackers would.

I made the simplifying assumption that if you target someone multiple times, and they have mirror images, you'd spread out the choices. So, each missile will target a different copy if possible, one of which may be the real creature. If you target with _more_ missiles than there are copies, it distributes them as evenly as possible, and it's possible some of the missiles will have no effect, because they both targeted the same mirror image.

This isn't a strict buff or nerf. It means that Mirror Image can provide protection from Magic Missile. But it also means that Magic Missile can shred off up to 5 images in one go.

The lack of interaction was Atari bug 161.

I didn't do KotB magic missile because the script wasn't in there already. Should I copy it over from the Co8 directory? It kind of looks like the one in the module is unmodified from ToEE (so, has a bunch of special cases for particular NPCs that I guess are no longer necessary in T+, and might not even exist in KotB).

### DIFF
--- a/TemplePlus/python/python_dice.cpp
+++ b/TemplePlus/python/python_dice.cpp
@@ -78,7 +78,7 @@ int PyDice_Init(PyObject *obj, PyObject *args, PyObject *kwds) {
 	} else {
 		self->bonus = 0; // Initialize since it's optional
 		// Form: PyDice(1, 2[, 3]) for 1d2+3 etc.
-		if (!PyArg_ParseTuple(args, "ii|:PyDice", &self->number, &self->size, &self->bonus)) {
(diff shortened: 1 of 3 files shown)


doug1234:
Nice!  

You should grab the kotb script and update that also.  

Why did you make it stricter rules only?  If its an Atari bug it seems like it should be fixed for all.

dolio:
I made it strict rules because while I do think it is the way it's supposed to behave, I could see people preferring the vanilla way. There are a few Atari bugs that are rules judgments that I'm not sure are 100% either way.

PR #786, opened 2024-08-08
Update char_class_utils.py fixed a "elif" when it should be "if"

nwalshgit:
spells_known = {
            1: [6, 3, 2],  # at first level we know 6 cantrips, 3 first and 2 second level spells
            2: [6, 5, 4, 1], # at second level we know 6 cantrips, 5 first, 4 second and 1 third level spell
        }
        METHOD FAILS
        caster_level=3, spell_level=1  ==> 3 but should be 2
        caster_level=3, spell_level=3  ==> 0 but should be 1

### DIFF
--- a/tpdatasrc/tpgamefiles/rules/char_class/char_class_utils.py
+++ b/tpdatasrc/tpgamefiles/rules/char_class/char_class_utils.py
@@ -85,7 +85,7 @@ def GetSpellsKnownAddedCount( spells_known, spellListLvl, spellLvl):
 				prevNumKnown = q[spellLvl]
 			else:
 				prevNumKnown = 0
-		elif (p > highestEntry and p < spellListLvl):
+		if (p > highestEntry and p < spellListLvl):
 			highestEntry = p
 	# print "newNumKnown is " + str(newNumKnown) + "prevNumKnown is: " + str(prevNumKnown)
 	if (newNumKnown == -999):


doug1234:
Change looks good.  Can you provide a quick summary of what this fixes from a player perspective for the change log?

nwalshgit (commented):
Fixes spell level progression when extrapolated beyond spells_known definition.

def test_func():
    spells_known = {
        0: [],  # No spells before taking a level
        1: [3],  # At first level of class, we get three cantrips
        2: [3, 1],  # At second level of class, we get one first level spell
        3: [3, 3],  # At third level of class, we have three cantrips and three first level spells
        # Currently we assume that the progression continues based on delta of last two levels
        # so if we added two first levels spells when we last gained a level, then we would keep doing that
    }
    assert GetSpellsKnownAddedCount(spells_known, 0, 0) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 0, 1) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 0, 2) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 1, 0) == 3, ""
    assert GetSpellsKnownAddedCount(spells_known, 1, 1) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 1, 2) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 2, 0) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 2, 1) == 1, ""
    assert GetSpellsKnownAddedCount(spells_known, 2, 2) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 3, 0) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 3, 1) == 2, ""
    assert GetSpellsKnownAddedCount(spells_known, 3, 2) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 3, 3) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 4, 0) == 0, ""
    assert GetSpellsKnownAddedCount(spells_known, 4, 1) == 2, "We should get 2 first level, like we did when we went from 1->3. Not just look at total we had at level 1."
    assert GetSpellsKnownAddedCount(spells_known, 4, 2) == 0, ""

doug1234:
Thanks for that info!

PR #796, opened 2024-11-09
Add a scale factor for the console/dm text

dolio:
This adds a (currently file-only) configuration value that controls the font scaling factor in the imgui stuff implementing the top console and dungeon master UI.

Most stuff in the game is (I think) drawn at the resolution set in the Temple+ config and then scaled to your screen's actual resolution, so it gets scaled. However, the imgui stuff gets drawn at the actual screen resolution. So, if you have something like a 4K monitor, it ends up pretty small. This lets you put e.g. `dmGuiScale = 2.0` in the .ini file to get a more legible result. Or you could scale it even bigger if you want.

I could add it to the configuration app, but I'm not sure where to put it in the UI. Right now you just have to edit the ini file by hand. After you run Temple+ once the default value (1.0) will be in the file, so you don't have to guess the name, but it might not be obvious that it's there. Let me know if you have ideas of where would be good to put the configuration.

Also, if anyone knows a better way to use the scaling in the imgui, let me know. Right now it's just calling the per-window scale factor function on every window. I tried setting the global font scale factor at the start, but it didn't seem to work. Maybe I didn't do it quite right, though. Not sure.

### DIFF
--- a/TemplePlus/config/config.cpp
+++ b/TemplePlus/config/config.cpp
@@ -140,6 +140,7 @@ static ConfigSetting configSettings[] = {
 	CONF_STRING(defaultModule),
 	CONF_INT(windowWidth),
 	CONF_INT(windowHeight),
+	CONF_DOUBLE(dmGuiScale),
 	CONF_BOOL(enlargeDialogFonts),
 	CONF_BOOL(windowedLockCursor),
 	CONF_BOOL(dungeonMaster),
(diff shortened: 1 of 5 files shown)


doug1234:
Another nice change!  I am going to merge your stuff into features soon.  Hoping to get a beta of the next version ready some time in December...

PR #797, opened 2024-11-09
Reimplement Raise Dead (and related) logic

dolio:
This PR reimplements the resurrection logic directly in C++, and replaces the original dll function for it.

The immediate benefit is that Raise Dead will work on Aasimars and Tieflings, because there is added logic to handle that case. This is similar to the special case for Darley in the original code.

But also, it implements the correct checks for Resurrection and True Resurrection spells. The original code only handled Raise Dead and St. Cuthbert cases, because those were the only ones in the game. No one's actually added those spells, but it doesn't seem like they would have worked properly if they had anyway.

Another thing I've added so far is an actual negative level from Raise Dead and Resurrection. The original game just left that out, so I put it in only on strict rules. It was doing the constitution modification for level 1 people, oddly enough.

I have a few other things to try to fix up in this area. But I thought I'd start a draft PR as a preview. Stuff to come includes:

- There are duplicate "resurrection successful" messages when someone is resurrected from a spell, because both the condition and the implementation float the lines.
- I'm not sure the resurrection function cures all the conditions it should (like poison and disease). Maybe they automatically cancel when you die, though, so I'll check that out.
- The level drain aspect seems a little funky, so I might need to look into that.

BTW, it doesn't seem like the existing fix that was added for Raise Dead works. I think a misspelling (maybe because the Python code actually spells 'Aasimar correctly') just causes the spell to not work at all. The quick fix can be removed if/when this gets merged in, though.


dolio:
BTW, does anyone know why the Raise Dead spell script sets `global_vars[752]` to the XP value it's going to reduce the target to? It doesn't seem necessary, and that global variable is used in a different script for an unrelated purpose.

anatoliy-savchak:
Nice. But I do think all of that should be in Python. No business logic / rules in cpp.

doug1234:
Very nice change.  I have been meaning to replace this code like you did but haven't gotten around to it.  For some reason I thought that global variable was used in the C++ code somewhere.  Is that not the case?

I think the ideal way to handle the aasimars and tieflings is to add a native outsider subtype rather than hard coding.  But I am OK with it the way it is.

dolio:
Oh, I can add a native subtype. I agree that'd be better. The space seems like it's getting slim, but it's probably worth it.

I don't see in the temple code where that variable would be read in conjunction with resurrection stuff. There is one line in the Raise Dead script that does:

```
	target_item.obj.object_script_execute( target_item.obj, 18 )
```

And I don't know what that script does, so I guess it's possible that it's read there. I'm unsure what it'd be doing, though.

dolio:
Here's another question...

_Why_ does the Raise Dead script set the XP value at all? The original `ResurrectApplyPenalties` sets the critter's XP value already. So does applying a permanent drained level. And the logic in the Raise Dead script can actually set it to the wrong thing if you have drained levels, because the level check it's calling doesn't seem to take them into account (so it resets your XP to a level below your _maximum_ level, which might even increase it). I'm not sure it works with level adjustment, either.

It seems to have been added by Co8 at some point, because I don't see it in the script extracted from the original dats. Does anyone know what it was trying to work around?

doug1234:
I suggest asking on the co8 forum and see if anyone there knows about the global variable.

dolio:
Here's the info from Co8:

- The global variable in the XP stuff is probably incidental. It matches the style of the person who probably originally fixed the XP problem, but I don't think there's any reason to use a global variable for it instead of a local variable. In any case, I just eliminated it and made the reimplementation actually set the right XP value (in the relevant case).
- Object script 18 is the resurrection script, which informs the game that the target has been resurrected. The example given was that if Meleny dies, Filiken will be angry and attack you. And in the original game, resurrecting her wouldn't fix that. I'll probably add a comment about that. Maybe it should also be moved into to resurrection function, because I'm not sure all the places it gets called will also call the object script. In any case, it has nothing to do with the global XP variable.

I'll probably wrap this up soon. It's not perfect, because `Perm Negative Level` actually isn't the exact right thing to be using when raising someone. However, I think fixing that properly requires a bit more sophisticated rework of level drain stuff that is probably better handled in a separate PR.

dolio:
FYI, I realized that making the XP stuff work correctly is a little more complicated than I thought.

When I tried making it use the ECL calculation, I noticed that it wasn't taking negative levels into account, because that function doesn't use the dispatch to take level drain into account. But, then I realized that that wouldn't be quite correct anyway, because that dispatch includes modifiers for (temporary) negative levels. So, if a creature had any of those, their XP would be drained _more_ than it should (and I think the existing level drain replacement would be affected by this).

So, I've been working on adding a way to filter out certain types of level drain when doing the dispatch, so that you can calculate what a character's _permanent_ (relatively speaking) drained ECL is. Then this can be used to set the correct XP value when applying the permanent drain. I also included a third category for levels-drained-by-resurrection, which will be necessary to properly calculate what stuff like Restoration does.

But, I don't think I can get it all done tonight, and I'm not going to be able to work on it for the next couple weeks, so this PR will have to remain on hold for the time being.

dolio:
Okay, here's a summary of the latest modifications:

- The `GetLevel` dispatch now may have non-zero flags which indicate which sorts of negative levels you want to _omit_. By default this is `0` (omit nothing). The negative level conditions check this and skip their effects appropriately.
- There's a new function for querying the character's effective level _including_ level drains. By default this is set up to include permanent drains, because that's usually why you'd be calling it instead of the existing ECL function.
- Permanent negative levels and non-strict resurrections reset the character's XP based on the drained ECL, which prevents it from subtracting too much XP
- The threshold for removing a permanent negative level based on XP gain is now set to the target ECL, because before it was getting removed too early based on just the hit dice.

doug1234:
This is coming along well!  What is left to do?

dolio:
I think the only thing left from my original list is the doubled floating messages. I need to decide which function shouldn't be producing one, because right now both the core resurrection function and the condition-add (I think) functions for the spells do.

doug1234:
OK cool.  I am thinking about putting out a beta at some point.  Either in a couple of weeks or at the beginning of the new year.

dolio:
Okay. I think this is about ready.

- I removed the float message from the resurrection function, so only the ones from spells will occur. There are some scripts that call the python resurrect function, which won't automatically float the lines anymore. I don't think it's a big loss. Basically, in the Cuthbert/Iuz scene, the messages won't be floated above all the targets. The python function returns an indicator of success, so the scripts could float messages if desired.
- I replaced the implementation of the Jaroo spell permanency fix. Before it was killing the player, having Jaroo cast reincarnate, then restoring the character's XP. With the strict rule mode, that was leaving the character with a drained level. Instead, it now does sound/animation stuff so it seems like he's casting reincarnate, but uses the Cuthbert resurrection, which automatically has no penalties.
- I removed the custom XP stuff from the reincarnate spell script, and replaced the `18` with the `san_resurrect` constant, like in Raise Dead.

dolio:
BTW, is there a changelog you want me to put stuff in somewhere?

doug1234:
Change log is here:  https://github.com/GrognardsFromHell/TemplePlus/wiki/Changelog under the 1.96 section.  Can you edit it?

dolio:
BTW, working on one last thing. There's supposed to be a 50% chance of losing each of your remaining spells when you get raised. Wasn't implemented originally, but I should be able to add it, I think.

dolio:
Okay, I think I'm done with all my tweaks.

dolio:
BTW, note that the strict rules change adding a level drain on raise doesn't make it easier to get your XP back when dying. (Currently) Restoration doesn't check if you have negative levels, it just restores your XP if it's below what would be needed for your level ups. So it would already (erroneously) give you back XP after you died.

PR #800, opened 2024-12-08
Clairify one handed wield with a comment.

doug1234:
Currently the option only displays when the character has a buckler which seems like a bug.  This is due to is buckler returning false when the input is null.  Now it should display whenever a character has no shield.  Should it only display for medium sized or larger weapons?


doug1234:
It is actually correct.  Just add a comment for future reference.

PR #815, opened 2025-01-06
Removed an extra null character added in local_to_ucs2 for isssue #810.

### LINKED ISSUE #810: Weird font issues under wine/linux
I've been trying to look into a font issue that I started having a few months back (after an extended hiatus). I don't think it used to happen, but I'm not certain. Here's a picture of the problem:

image

Obviously the big square isn't supposed to be there. It doesn't happen with every font. I think those numbers are drawn with `ARIAL_BOLD_24`, but `ARIAL_BOLD_10` doesn't have the same problem. I haven't tried all variations.

My question for other folks is, does this happen on other systems? I'm running Temple+ in Wine on Linux, and I was assuming it doesn't happen on Windows (because someone would have mentioned it already). Has anyone else been seeing this behavior and just living with it?

---

Side note: If I fiddle around with the font mapping files, I can change it to another font. If I change it to Junicode like the entry for Priory, it looks like this:

image

Question mark instead of a box. I think in each case that's the font's symbol for an unknown glyph. However, it only shows up when drawing at certain sizes, which is pretty weird.

doug1234:
Closes #810

### DIFF
--- a/Infrastructure/stringutil.cpp
+++ b/Infrastructure/stringutil.cpp
@@ -22,11 +22,10 @@ std::string ucs2_to_local(const std::wstring&s) {
 
 std::wstring local_to_ucs2(const std::string &s) {
 
-	auto slength = (int)s.length() + 1;
-	auto len = MultiByteToWideChar(CP_ACP, 0, s.c_str(), slength, nullptr, 0);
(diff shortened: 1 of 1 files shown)


dolio:
This change is essentially what I had arrived at myself, so it looks good to me.

Issue #831, opened 2025-01-15
My Characters Can No Longer Cast Flame Strike


Legion50:
This is a save issue instead of a bug, I'm hoping you guys can look and see what went wrong here. I have a Cleric, a Druid, and a Warmage of appropriate level, and they have all been using Flame Strike just fine... until now. I first ran into this problem while fighting the Hextorites at the Temple, and then I was ambushed by Robin of Locksley and her crew. Even after resting, none of my characters are able to cast the spell; they do an animation (the _wrong_ one, at that) and nothing happens. Thoughts? (Use IWD portraits)
[Flame Strike Issue.zip](https://github.com/user-attachments/files/18420105/Flame.Strike.Issue.zip)

dolio:
Well, something like that could explain it if it were making your characters deaf (which can cause spell fizzling). It seems like all your guys are properly immune, though.

Legion50:
Hmmm... if you were to cast Divine Favor/Divine Power, it looks like that. It's SUPER weird. The spell isn't being miscast (as part of a mechanic), the spell is literally just failing to execute correctly.

Legion50:
Hello, I have once again returned from the dead. Closing this because this issue is no longer reproducible at this time. Cheers.

PR #832, opened 2025-01-19
Use ag_animate for dismiss animation

dolio:
This is a quick fix for some issues I noticed with dismissing certain spells. For instance, I noticed that when you dismiss Silence, it actually restarts the spell.

The issue goes back to when I added spell dismissal animations. I mistakenly thought it was safe to use one of the spell casting animation goals for that. But, the problem is that the animation sequences actually contain callbacks the execute game logic in response to certain portions of the animation completing. So, using the `ag_throw_spell_w_cast_anim` sequence will actually trigger some of the python calls.

In most cases this actually turns out not to be a problem, which I guess is why I didn't notice. For most spells, the dismiss event ends the spell, and the script triggers don't resume it. _However_, there are some spells that lack the right dismiss handling. In vanilla ToEE, that means that when you click `Dismiss Spell`, nothing happens. Grease and Silence are examples. Temple+ fixed this by making the `Dismiss` handler also call some of the spell ending functions. But apparently that isn't good enough because the wrong animation sequence can result in the spell being refreshed.

So, this switches to making dismiss use the `ag_animate` goal, which is used in e.g. `anim_goal_push` in python. I think it just plays the animation without any of the game logic.

Sorry I didn't catch this earlier. I guess you could consider it something caught in 'beta testing' the next release. :)

### DIFF
--- a/TemplePlus/animgoals/anim.cpp
+++ b/TemplePlus/animgoals/anim.cpp
@@ -506,10 +506,13 @@ BOOL AnimSystem::PushSpellCast(SpellPacketBody & spellPkt, objHndl item)
 BOOL AnimSystem::PushSpellDismiss(SpellPacketBody & pkt)
 {
     AnimSlotGoalStackEntry goalData;
-    if (!goalData.InitWithInterrupt(pkt.caster, ag_throw_spell_w_cast_anim))
+    if (!goalData.InitWithInterrupt(pkt.caster, ag_animate))
         return FALSE;
 
     SetGoalDataForSpellPacket(pkt, goalData, true, true);
+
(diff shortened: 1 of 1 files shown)


doug1234:
Good find.  I will merge soon.  Does this have anything do do with issue #831?

dolio:
No, not that I know of. I haven't actually  reproduced the problem there.

I found this when I was working on getting an old branch with Silence related stuff up to date, and seeing what all there was left to do. I thought I had somehow broken dismissing it in that branch. But then I realized it wasn't that branch that was the problem at all, and it wasn't just Silence.

PR #834, opened 2025-02-18
Rework armor check penalties and max dex bonus to be less convoluted

### LINKED ISSUE #803: Inappropriate Armor Check Penalties
The following armors have inappropriate armor check penalties being applied when they should be 0, resulting in penalties to Hide, Move Silently, Sleight of Hand, and Tumble.

- Darkwood Buckler
- Large Darkwood Shield
- Large Mithral Shield
- Mithral Shield
- Wooden Elvish Shield

This list may not be exhaustive, I only tested equipment that my party would be interested in using at the time.

dolio:
Lining up some PRs of stuff I've been working on for after the imminent release.

This one reworks armor check penalties and max dex AC bonuses and fixes some associated items.

- I added a `DispatchForWearable` that combines the way these worked with an item query.
    - I think the reason that e.g. `GetArmorCheckPenalty` was never used is that the masterwork condition is an _item condition_ that gets applied to a creature when it wears the armor. But, the existing dispatcher would look for the parent of the armor, and dispatch on it. So, including an armor check callback to adjust the penalty would only display the right value while it was being worn.
    - My version dispatches against the parent if applicable, and does an item query if that doesn't work, or if the item isn't worn. So it will always include the item condition. I did it this way so that if people want to make feats or other features that affect the armor check penalty, they can (some kind of armor mastery or something). The stats you see in the UI will be relative to whoever is holding to the armor, if they have some creature condition that affects it.
- I made both armor check penalty and max dex AC use this dispatcher. I did that because I noticed both were adjusted in similar ways to include an extra python query. But with my changes the python query is no longer necessary, because the base dispatcher is bonus list based.
- I made `Armor Nimbleness` use the basic query, and added an extension to the `Armor Masterwork` condition that adjusts the armor check penalty
- I changed the encumbrance conditions to _not_ respond to these queries, and instead directly put caps on Dex AC. At first I just did this so I could use the dispatcher to get the description values. But, it also fixes the fact that being encumbered would make it look like your _armor_ had a lower max dex bonus than it actually did (because the `Armor Bonus` condition was doing the AC capping). Now the AC breakdown shows that your Dex AC is being capped by weight encumbrance, not armor.
- I also noticed that the encumbrance caps were wrong. All the caps were 3, when it should be 1 for heavy, and 0 for overburdened (and the last also removes your dodge AC and allows you to be sneak attacked). So I fixed those.
- I replaced all the skill bonus callbacks in `Armor Masterwork` with no-ops because they're no longer necessary.
- I removed the circumstance skill penalties from the Co8 masterwork items that were trying to offset the masterwork bonuses.
- I fixed masterwork gladiator armor having its armor check double reduced (the base was set lower, _and_ it was masterwork).
- Made body armor check penalty not stack with 'armor check' penalty from weight encumbrance
- Added str/dex ability check penalties for body armor when non-proficient (they were already on shields)
- Changed non-proficiency armor check penalty to Survival to be Use Rope instead (no idea why it was survival on both armor and shields; that's a wisdom skill)
- Added improved in-game help text

Fixes #803


dolio:
So, here's more of a high level question I have with regard to this, in case anyone wants to chime in...

Why does masterwork stuff work this way? Is there a benefit? Every masterwork armore has a separate proto right now, where the armor check penalty of the _base_ item is given, and then the masterwork condition adjusts it by +1 (now working correctly, but previously with a bunch of idiosyncratic bonuses). But the proto could have just had the stats of the masterwork version of the item instead. For masterwork weapons you need a condition to add the +1 to hit bonus. But for armor you can just put different values in the fields.

Is there a point? It would make some sense if there were _just_ protos for the mundane items, and then you could easily create masterwork versions programmatically by applying the one item condition. But that doesn't seem to be done (except maybe via copy-paste in the proto). Do some of the other modules do it?

doug1234:
Good stuff! I will merge after we finalize the current version.  Sadly I haven't heard anything from beta testers.  

I have no idea why masterwork armor is the way it is sadly.

doug1234:
Did you intend to keep this one in draft?

dolio:
I think it was a draft in anticipation of having to potentially having to fix after the release.

PR #835, opened 2025-02-24
Use base radius for clearance in pathfinding

dolio:
Allows PCs through doors they'd normally be able to get through as in the base game.

### DIFF
--- a/TemplePlus/obj.cpp
+++ b/TemplePlus/obj.cpp
@@ -440,7 +440,7 @@ float AdjustRadiusSize(float baseRadius, int baseSize, int curSize)
 	return radius;
 }
 
-float Objects::GetRadius(objHndl handle)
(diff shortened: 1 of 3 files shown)


doug1234:
Change looks good.  Do you think I should put out another beta?

dolio:
I think it's not strictly necessary. But if it's easy, people who are trying the current beta might appreciate being able to walk through doors while enlarged.

DudeMcDude:
> Change looks good. Do you think I should put out another beta?



> I think it's not strictly necessary. But if it's easy, people who are trying the current beta might appreciate being able to walk through doors while enlarged.

How well tested is it? These subtle PF changes can be tricky from what I recall.

doug1234:
Sounds like another beta is needed then.

PR #838, opened 2025-02-28
Various poison related fixes

dolio:
This PR contains various poison related fixes. I did most of them a while ago, but I recently cleaned them up and updated the branch.

- Implemented actual delayed poison. In the existing game, delay poison would just cause you to just not get poisoned. Now, with strict rules enabled, you will receive a condition that waits until the delay spell wears off, and then applies the poison. Similar situation with the secondary damage, in case the delay is cast after the primary damage.
- Reduced Neutralize Poison duration to 10 min/level, instead of 1 hr/level
- Changed the way Vrock spores work. They're not a spell, so I made the 'spell' script to just apply the condition and end. I also changed around the immunities and such to match the SRD on strict rules. For instance, delay poison only delays the effects (that this can delay the condition varying amounts for different targets was the reason I didn't want it to act like an ongoing spell).
- Made Delay Poison remove Hezrou stench, per the rules
- Made an option to avoid copying over the poison specs from the original dll, so that they'll be loaded from `poisons.json`. Also fixed the file path to `poisons.json`. This allows fixing poisons that were wrong on the original game.
- Fixed some of the poison specs (like the adder poison, which fixes Atari bug 274)

I think some of the fixes to the 'has condition' callback in here might no longer be necessary, because it's been fixed in a more general way since I wrote this. I'll take a look if they can be rolled back at some point.


doug1234:
Did you intend to keep this one in draft?

dolio:
I think this is almost ready to go. Let me check the condition part I mentioned, though.

PR #840, opened 2025-03-08
Replace racial ability modifier function

### LINKED ISSUE #839: Odd behavior of racial ability adjustments for low scores
So, I was checking to make sure that some stuff I wrote for the con penalty when getting raised at level 1 didn't do something weird to characters with a racial con penalty, and came across a different weird bug instead.

It's clear that the game is trying to make all racial penalties bottom out at 3. If you roll a 3, and then choose a race with a -2 penalty, it won't actually apply the -2. This intention isn't actually correct, because as far as I can tell, only intelligence has an exception like this (because you need 3 int to speak, and gain class levels, I think). But also, the math isn't even correct. If your unadjusted stat is 4, then it will accidentally cap the penalty to -1. But if your unadjusted stat is 3, it will still give the -2, and if it's _below_ 3 it will give a _bigger_ penalty than normal.

I was going to fix this, but my question is how I should do so. Should I try to test for strict rules, and if they're not enabled actually implement the game's intention of clamping base ability scores at 3? Or should I just implement the actual rules, because essentially no one is going to have scores that low anyway?

dolio:
This replaces the callback for racial ability modifiers, with a few effects.

- The old callback had some wonky math going on for very low scores. It tried to boost things up to 3 in the case a penalty was being applied, but instead it was sometimes applying _larger_ penalties. The new callback just implements the rule of a racial penalty to intelligence not modifying your score to a value below 3, from the SRD.
- I made the test look at the critter's stored int32 value to determine whether the racial penalty is taking you below 3. This means that it doesn't matter what order conditions get added in, the behavior will be the same. Previously it was just adding up whatever was already in the bonus list to see what to do.
- I made negative modifiers display "Racial Penalty" in the info screen, instead of "Racial Bonus"

Fixes #839

### DIFF
--- a/TemplePlus/general_condition_fixes.cpp
+++ b/TemplePlus/general_condition_fixes.cpp
@@ -27,6 +27,7 @@ public:
 	static int WeaponKeenCritHitRange(DispatcherCallbackArgs args);
 	static int ImprovedCriticalGetCritThreatRange(DispatcherCallbackArgs args);
 	
+	static int RacialAbilityAdjustment(DispatcherCallbackArgs args);
 
 	void apply() override {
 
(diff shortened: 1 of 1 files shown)


doug1234:
This change looks good.  Need to get the current beta finished up and get this merged I think.

dolio:
Yeah, there's no rush.

Honestly, the thing I'm anticipating most is you merging the features back into trunk so that the universal dmgui scaling is on. :stuck_out_tongue:

PR #844, opened 2025-03-23
Add debugging/default to physical stat functions

dolio:
When going to debug the physical stat functions, I also had it just return 0 when the handle is null, and that seems to fix the crash when (some) NPCs are killed.

What seems to happen is that the game prints some NPCs' dying words by calling the function for speaking to someone with a null `speakingTo`. This eventually tries to look up the gender of `speakingTo` to get the right dialogue/sound, which would bomb with a null handle (direct getInt32 on a null object). There seem to be other scenarios where `speakingTo` is null, too, so it isn't just for dying creatures.

There was already a replacement adding a null check to the DLL physical stat function, but I didn't notice it when rewriting the functions directly, so I didn't add null checks, and bypassed the existing null check when calling from `StatLevelGet(Base)`.

I can remove the debug logging if we think that it's not really an erroneous case.

### DIFF
--- a/TemplePlus/gamesystems/d20/d20stats.cpp
+++ b/TemplePlus/gamesystems/d20/d20stats.cpp
@@ -308,6 +308,11 @@ int D20StatsSystem::GetPsiStatBase(const objHndl & handle, Stat stat, int statAr
 
 int D20StatsSystem::GetPhysicalStatBase(const objHndl & handle, Stat stat) const
 {
+	if (!handle) {
+		logger->debug("GetPhysicalStatBase called with null handle for stat {}", stat);
+		return 0;
+	}
+
 	switch(stat){
 	case stat_ac:
 		return critterSys.GetArmorClass(handle);
@@ -356,6 +361,11 @@ int D20StatsSystem::GetPhysicalStatLevel(const objHndl & handle, Stat stat) cons
 	auto sizeDiff = curSize - baseSize;
 
 	int base = 0;
+
+	if (!handle) {
+		logger->debug("GetPhysicalStatLevel called with null handle for stat {}", stat);
+		return 0;
+	}
 
 	// TODO: polymorph
 	switch(stat) {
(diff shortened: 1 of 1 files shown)


doug1234:
It seems to me it wasn't originally intended to be called with a null handle.  I didn't see any protection in the decompiled code unless I am missing something.  I can merge it in as is and we can see how it goes.

dolio:
I'm pretty sure the way this happens for dying is...

1. Getting to -10 HP causes `Kill(critter)` in one of the hp changed callbacks, or you call the equivalent elsewhere
2. `Kill(critter, killer)` executes `DoOnDeathScripts(critter, killer)`, and `killer` defaults to null
3. `DoOnDeathScripts(critter, killer)` exectues `critter`'s `san_dying` with `killer` as an argument
4. `san_dying` calls `ShowTextBubble` with `killer` as `speakingTo`
5. `ShowTextBubble` looks up the gender of `speakingTo` without checking if it's null

Allowing `Kill(critter)` with a default null is a Temple+ addition, obviously, but there's a bunch of `CritterKill(critter, 0)` in the original DLL, including in one of the hp changed callbacks. It works in vanilla ToEE, but I don't understand how.

I don't know for sure that step 4 is what happens directly, because I don't know where `san_dying` is. But if you do `furnok.execute_object_script(OBJ_HANDLE_NULL, san_dying)` on the release, you get the crash. And I know that `ShowTextBubble` gets called with null, and it does a stat query on the null object.

dolio:
Here's a theory. This works in vanilla ToEE because running `GetInt32` on the null object handle doesn't crash. It's just looks up whatever the value is for the lowest thing in the master object array (or something like that). And in this case it doesn't actually matter what the result is, because the NPC dialogue is the same regardless. So Temple+ originally 'broke' this scenario by crashing instead of giving you arbitrary results for the null object handle.

DudeMcDude:
Yes, in vanilla ToEE getX functions just smooth over null handles and return 0. 
In T+ we made it throw (from within the C code) to catch bugs. For the python API functions it does return 0, IIRC because there are many violations in the scripts.

I'm not sure an extra debug line is warranted, but I guess it's ok if it doesn't result in log spam.

dolio:
Okay.

Removed the logging, just in case there are other scenarios where it gets called this way more often.

DudeMcDude:
Yeah, should probably release a hotfix update.

doug1234:
I am going to put out a hot fix this evening.

doug1234:
I am now convince this is a good fix.  Merging.

PR #849, opened 2025-03-27
Fix new base classes not getting starting equipment

### LINKED ISSUE #848: Swashbuckler doesn't get free stuff
Minor nitpick but Swashbuckler doesn't get  the weapon he gets proficiency in like with Exotic Weapon Proficiency (Spike Chain) and buckler unlike other classes with Weapon Proficiency or Weapon Focus. Is it because he automatically has Weapon Finesse although I doubt it since he spawns naked with nothing.

Edit: Just found out that Swashbucklers don't get Shield proficiencies.

dolio:
The function that loops over classes and calls the python `pc_start` function was only checking for vanilla classes. I changed it to loop over all the base classes instead, because `pc_start` is how you get bonus items for exotic weapon proficiencies and focus.

I also augmented the Co8 `start_equipment.mes` to give farmer clothes to the new classes. It probably doesn't matter because I assume everyone throws those away, but it's consistent at least.

In other modules they still won't get a starting loadout, but we don't seem to have files for those right now. KoTB has a shop map if I recall correctly, so it probably doesn't matter. Not sure about other mods. Vanilla gives actual loadouts for each class, but I don't actually know where to put stuff for only the original ToEE module.

Fixes #848

### DIFF
--- a/TemplePlus/python/python_integration_obj.cpp
+++ b/TemplePlus/python/python_integration_obj.cpp
@@ -74,15 +74,17 @@ static void PcStart(objHndl pc) {
 	inventory.Clear(pc, FALSE);
 	
 	// This checks that the PC has at least one level in any of the classes
-	auto stat = d20ClassSys.vanillaClassEnums[0];
-	auto classIndex = 0;
-	while (classIndex <= VANILLA_NUM_CLASSES) {
-		stat = d20ClassSys.vanillaClassEnums[classIndex];
+	int classIndex = 0;
+	for (auto classEnum : d20ClassSys.baseClassEnums) {
+		auto stat = static_cast<Stat>(classEnum);
+
 		if (objects.StatLevelGet(pc, stat) > 0) break;
+
 		classIndex++;
-		if (classIndex >= VANILLA_NUM_CLASSES)
-			return;
 	}
+
+	// couldn't find class
+	if (classIndex > d20ClassSys.baseClassEnums.size()) return;
 
 	try {
 		auto content(MesFile::ParseFile("rules\\start_equipment.mes"));
--- a/TemplePlus/python/python_integration_obj.cpp
+++ b/TemplePlus/python/python_integration_obj.cpp
@@ -86,6 +86,9 @@ static void PcStart(objHndl pc) {
 	// couldn't find class
 	if (classIndex > d20ClassSys.baseClassEnums.size()) return;
 
+	// offset by 1 for tutorial inventory
+	if (classIndex >= VANILLA_NUM_CLASSES) classIndex++;
+
 	try {
 		auto content(MesFile::ParseFile("rules\\start_equipment.mes"));
 		
--- a/tpdatasrc/co8infra/rules/start_equipment.mes
+++ b/tpdatasrc/co8infra/rules/start_equipment.mes
@@ -42,20 +42,23 @@
 // wizard 
 {10}{6142} 
 
-// favored soul
-{11}{6142}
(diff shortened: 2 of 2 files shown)


doug1234:
Is something similar needed for KotB?

dolio:
I think with this merged, a script will get called that gives the weapon focus/proficiency item, even in KotB. That script is not class specific, it just checks for feats.

What won't happen is the class-based initial load out. In Co8, every load out (except the tutorial) is farm clothing, which I assume everyone just drops. If memory serves, KotB is the same, so it probably doesn't matter if the characters start in their underwear or farm clothing. I could try to see if it's just the same file as Co8 and add it, if you want.

I haven't played Paladin's Cove. Does that also start with the shop map and everyone wearing plain clothing?

I think the one that would actually be important is vanilla, where the file gives you your actual starting equipment. But I don't know if anyone still plays that (I know some things that could be fixed here aren't, because they're already fixed by using Co8). And like I said, I don't know where to put something that would apply only for vanilla. Does stuff in `co8...` override `tpgamefiles`? Or does it not work that way?

doug1234:
I don't think its worth looking into more.  

There is no shop map in paladin's cove.  Files from tpgamefiles.dat should be getting overridden by the co8 stuff.

DudeMcDude:
> I think the one that would actually be important is vanilla, where the file gives you your actual starting equipment. But I don't know if anyone still plays that (I know some things that could be fixed here aren't, because they're already fixed by using Co8). 

Yes, people still play vanilla. For example, non English versions, or even regular GoG players, or people who don't like mods (and are using T+ as a fixpack).

> And like I said, I don't know where to put something that would apply only for vanilla. Does stuff in `co8...` override `tpgamefiles`? Or does it not work that way?

Co8 as in normal Co8 or the T+ co8 fixes etc?
The hierarchy is specified here:  
https://github.com/GrognardsFromHell/TemplePlus/wiki/Files

dolio:
I meant `tpdatasrc/co8*`. We already have the starting equipment file there, so it should override if I add default equipment to `tpgamefiles`. I guess at that point it'd be necessary to add an override for KotB.

There's one discrepancy I can think of. Certain fixes that should apply to vanilla might include things like protos that wouldn't apply to a mod, because it's fixing something ToEE specific. So it seems like there could be situations where there's no appropriate place to put a fix that should apply to vanilla.

Also, a bunch of things on the Atari bug list are marked as fixed by Co8. Are those fixes also included in Temple+ for vanilla? Or maybe, but not systematically?

DudeMcDude:
> I meant `tpdatasrc/co8*`. We already have the starting equipment file there, so it should override if I add default equipment to `tpgamefiles`. I guess at that point it'd be necessary to add an override for KotB.

Yes, sounds right. 

> There's one discrepancy I can think of. Certain fixes that should apply to vanilla might include things like protos that wouldn't apply to a mod, because it's fixing something ToEE specific. So it seems like there could be situations where there's no appropriate place to put a fix that should apply to vanilla.

That could happen. In practice I think we only apply fixes to a few generic protos that are unlikely to be replaced (wolves, trolls and such). If it's going to be more extensive then yeah, maybe ToEE vanilla specific fixes should be supported.

> Also, a bunch of things on the Atari bug list are marked as fixed by Co8. Are those fixes also included in Temple+ for vanilla? Or maybe, but not systematically?

The DLL fixes from moebius and spellslinger are included. Otherwise, generally not. Especially stuff that's related to protos, MOBs, and dlg/scr fixes.

PR #851, opened 2025-03-29
Avoid dragging rolls during point buy.

dolio:
Intercepts mouse events that allow accessing the rolled stats during point buy. Left clicks on the rolled stat boxes are just dropped, and drag events onto the rolled stats are tweaked to appear outside the rolled stats when delegating to the DLL, so that the drag is just aborted.

Fixes Atari bug 338.

As an additional effect, the bug where you can double click on one of the rolled stats to get your roll count as a stat during point buy is also fixed.


DudeMcDude:
Err, this is a personal favorite of mine - I always deemed this too funny and iconic to fix...

dolio:
I figured that might be the case. :)

Issue #852, opened 2025-03-29
Grease spell lingering effects, stirges not attacking, sneak attack issues


ALFAIthildur:
1. Grease spell effect lingers on party members long after duration should have expired after combat ends
2. stirges do not attack
3. party rogues are able to sneak attack enemies (ambush assassin, Ranos Davil) that should not be sneak attackable due to having much higher rogue lvls and uncanny dodge

dolio:
For 3, I'm pretty sure improved uncanny dodge is fine. The problem is that various things that enable sneak attack don't check for normal uncanny dodge. So for instance, if you're flatfooted, your sneak attack status is just set to 1. It only looks at uncanny dodge for the AC check (even though that's technically what determines whether you can be sneak attacked). I think invisibility/blindness probably work the same way.

I also noticed there's some other fishiness w/r/t rogue features. So I might do a pass that fixes up the above and the other stuff.

PR #855, opened 2025-04-09
Fix some errors related to sneak attack and uncanny dodge

dolio:
This fixes a few issues:

1. The flat-footed condition was unconditionally allowing sneak attacks. Uncanny dodge allows you to retain your dexterity AC bonus while flat-footed, and losing that is the scenario that allows sneak attacks. This fixes part 3 of #852.
2. Uncanny dodge also lets you keep your dexterity AC against opponents you can't see. So does blind-fight in melee. So, I've added a more specialized `CanSense` that incorporates this.
3. Uncanny dodge had some 3.0 cruft that added bonuses against traps above level 11. I think those turned into Trap Sense in 3.5, which is implemented directly in the class condition. So, I replaced those callbacks with no-ops.

I know of some other oddities related to this, but doing them properly involves other issues with invisibility being broken in various ways. So I'm submitting this standalone.

### DIFF
--- a/TemplePlus/condition.cpp
+++ b/TemplePlus/condition.cpp
@@ -1786,6 +1786,19 @@ int QuerySetReturnVal0(DispatcherCallbackArgs args)
 	return 0;
 };
 
+// Sets return_val to 1 if caller lacks feat in data1
+int QuerySet1IfLacksFeat(DispatcherCallbackArgs args)
+{
+	auto dispIo = dispatch.DispIoCheckIoType7(args.dispIO);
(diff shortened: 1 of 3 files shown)


dolio:
Note: the `CanSense` thing is not a super literal reading of the rules.  If you do that, then blind-fight and uncanny dodge only protect you from invisible people, not people you can't see for other reasons, like actually being blind. I think that's kind of stupid, and just a technical oversight that happened when they were rewriting things from 3.0 to 3.5. So, my intention is to implement things in the way that makes some amount of realistic sense, not some weird rules pedantry that doesn't.

PR #856, opened 2025-04-12
Add some spell information features

dolio:
This PR adds some features related to information about spells. So far, there isn't much that should actually affect gameplay that isn't already available. But it adds infrastructure that I'm planning to use in the future. Features are as follows:

- I made the file parser more robust, so that it checks for exact matches in e.g. the spell targeting, before checking for a prefix match. Before it was just looking for prefixes. There are targeting options named `Area` and `Area Or Obj`, but the parser would always parse the former when the latter was specified. This required some hard coded fixups for the spells that were supposed to use the second mode, and those are no longer necessary.
- Added a Universal spell school. There are a couple spells that belong to it, and every specialist wizard can use them, no matter their banned schools.
- Added the ability to specify the 'source' of a spell in terms of rulebooks and the like. This has a default based on the spell enum, but it can be overridden in the `rules` file for the spell with a `source:` line. This has the following effects.
   - Properly keeps track of which spells are core or not. Some core spells have been added with ids above 700-whatever, and Co8 added some non-core spells in the low range. I've fixed these, so having core-only enabled gives you the actual core spells.
   - I added a source selector to the config app, and you can choose specific sources to enable or disable. Personally, I find there are just too many spell compendium spells, so I'd kind of like them disabled, and this lets me do that. Right now it also means you can play with other non-core material, but only core spells, because I didn't add the information for all the other stuff (feats, classes, races, etc.).

I'm making this a draft for now, because I'm not sure I'm done with it. I'd appreciate some thoughts on stuff related to the individual sources, though. Is the GUI for it okay, or is there something better? I used a multi-select list box. I've noticed that it's hard to see what you have enabled if the box isn't focused, but that could just be because I run it in Wine.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,7 +114,7 @@ jobs:
       contents: read  # This is required for actions/checkout
     steps:
     - uses: actions/checkout@v4
-    - uses: actions/download-artifact@v3
+    - uses: actions/download-artifact@v4
       with:
         name: TemplePlusSymbols
     - name: Uncompress Symbol Files
--- a/Configurator/HouseRulesWindow.xaml
+++ b/Configurator/HouseRulesWindow.xaml
@@ -41,13 +41,25 @@
                 </Canvas>
             </TabItem>
             <TabItem Header="Extra Content">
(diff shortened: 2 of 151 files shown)


doug1234:
Is this one ready to go now?

dolio:
Yeah, I think this one is good to go, if you're satisfied with the config UI.

PR #863, opened 2025-04-30
Fixes/tweaks with reloading weapons

dolio:
This PR tightens up how reloading works.

1. Fixes a bug that caused the projectile animation to be skipped when making individual attacks as part of a full attack with crossbows. Ammo is only consumed at the end of the projectile animation, so this was allowing you to take all your attacks with a crossbow even without Rapid Reload (probably gave you infinite ammo, too).
2. On strict rules only, fixes some erroneous vanilla behavior
    - Reloading a heavy crossbow is a full round action, or a move action with Rapid Reload. Vanilla was move/free exactly like a light crossbow.
    - Makes slings a weapon that requires loading, costing a move action (with no Rapid Reload according to the SRD). I think all relevant reloading behavior has been implemented when this is turned on.

I didn't do anything with repeating or hand crossbows except leave some 'todo' notes. For the latter, it didn't look like there actually are any. For the former, you're supposed to load 5 bolts at a time, which would require additional machinery to track. Also, it seems like they're almost worthless if properly implemented. They require a feat, and they're still slower than light crossbow+Rapid Reload. I think even a heavy repeater is probably worse.

Crossbows have an item condition for manual reloading actions, but I decided to hook into the global radial builder for slings to avoid having to add proto overrides for every module.

### DIFF
--- a/TemplePlus/action_sequence.cpp
+++ b/TemplePlus/action_sequence.cpp
@@ -4069,7 +4069,7 @@ int ActnSeqReplacements::SeqRenderAttack(D20Actn* d20a, UiIntgameTurnbasedFlags 
 ActionErrorCode ActnSeqReplacements::ActionCostReload(D20Actn *d20, TurnBasedStatus *tbStat, ActionCostPacket *acp)
 {
 	// free action by default
-	acp->hourglassCost = 0;
+	acp->hourglassCost = static_cast<int>(ActionCostType::Null);
 
 	if (d20->d20Caf & D20CAF_NEED_ANIM_COMPLETED) return AEC_OK;
 	if (!combatSys.isCombatActive()) return AEC_OK;
@@ -4081,8 +4081,15 @@ ActionErrorCode ActnSeqReplacements::ActionCostReload(D20Actn *d20, TurnBasedSta
 	bool rapid = feats.HasFeatCountByClass(d20->d20APerformer, FEAT_RAPID_RELOAD);
 
 	// costs for reloading heavy/repeating crossbows based on strict rules status
-	int fastCost = config.stricterRulesEnforcement ? 1 : 0; // move or free
-	int slowCost = config.stricterRulesEnforcement ? 4 : 1; // full round or move
+	auto fastCost = ActionCostType::Null;
+	auto slowCost = ActionCostType::Move;
+
+	if (config.stricterRulesEnforcement) {
+		fastCost = ActionCostType::Move;
+		slowCost = ActionCostType::FullRound;
+	}
+
+	auto cost = ActionCostType::Null;
 
 	switch (objects.GetWeaponType(weapon))
 	{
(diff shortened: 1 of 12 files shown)


TemplePlus/action_sequence.cpp
```diff
@@ -229,10 +232,6 @@ ActionSequenceSystem::ActionSequenceSystem()
 	//rebase(GetRemainingMaxMoveLength, 0x1008B8A0);
 	//rebase(TrimPathToRemainingMoveLength, 0x1008B9A0);
 
-	rebase(_CrossBowSthgReload_1008E8A0, 0x1008E8A0);
-
-	rebase(ActionCostReload, 0x100903B0);
```
anatoliy-savchak, line 234: I thought we already have `stdAttackAddToSeq` reimplemented...

image

TemplePlus/action_sequence.cpp
```diff
@@ -4053,6 +4068,45 @@ int ActnSeqReplacements::SeqRenderAttack(D20Actn* d20a, UiIntgameTurnbasedFlags
...
+
+	auto weapon = inventory.ItemWornAt(d20->d20APerformer, EquipSlot::WeaponPrimary);
+	if (!weapon) return AEC_OK;
+
+	// TODO: rapid reload is actually a per-weapon feat per the SRD
+	bool rapid = feats.HasFeatCountByClass(d20->d20APerformer, FEAT_RAPID_RELOAD);
+
+	// costs for reloading heavy/repeating crossbows based on strict rules status
+	int fastCost = config.stricterRulesEnforcement ? 1 : 0; // move or free
+	int slowCost = config.stricterRulesEnforcement ? 4 : 1; // full round or move
```
anatoliy-savchak, line 4087: Costs instead of integers would be more readable as constants...

anatoliy-savchak:
Good change!
I was wondering why my NPC move after strike by heavy crossbow. thanks!

TemplePlus/action_sequence.cpp
dolio: We do. I'm not sure what the question is.

I replaceFunctioned `ActionCostReload` in case it's called in some DLL spot we haven't replaced. That isn't the only one, if I recall.

TemplePlus/action_sequence.cpp
dolio: Oh, and I can do this, I think. But `0` is actually called `Null` I think, which is a little weird. I think maybe because `Free` was the ToEE code name for what later came to be known as `Swift` actions (the ones that are quick like a free action, but you're only allowed to do one per round).

TemplePlus/action_sequence.cpp
anatoliy-savchak: This is where it's used:

image

anatoliy-savchak: Best assigned here for D20A_RELOAD, like this:

image

anatoliy-savchak: Therefor this function substitute hack could be removed altogether.

dolio: I guess now all the occurrences from the original DLL have been replaced, because I added the reload action to the new `d20Defs`. So yeah, the `replaceFunction` is unnecessary.

dolio:
Note: I didn't update any help. The original help actually describes the strict rules behavior, not what the game actually did.

doug1234:
This change looks good.  Is it ready to merge in?

dolio:
Yeah, I think it's ready.

PR #871, opened 2025-06-14
Apply skirmish damage to weaponlike spells

dolio:
These spells don't seem to dispatch `OnDealingDamage`, they dispatch a separate event. Sneak attack has been updated to listen for this event elsewhere, but skirmish wasn't.


doug1234:
Good find.

PR #872, opened 2025-06-27
Remove point buy cheating prevention

dolio:
This reverts the roll dragging fix without changing the factored out ui code that I added. I thought it might be nice to have that available without having to figure it out again.

I can't get the roll-count-as-stat bug to happen, neither with this branch nor with the last release. I thought it was just double clicking on one of the invisible stat boxes. Maybe it no longer works in wine or something. Since I can't make it happen with the last release, it seems unrelated to anything I changed. I don't understand what actually makes it happen, though.


DudeMcDude:
Thanks dolio!

PR #874, opened 2025-08-04
Fix memory bug.

### LINKED ISSUE #873: Temple+ Not Launching
I have had ToEE, Co8 New Content, and Temple+ installed for quite some time. I stopped playing some months ago, but never uninstalled anything. Recently decded to start playing again amd realized that I had Temple+, so I went to use that to play, but nothing happened.  It will not launch. I have uninstalled everything ToEE related, game, Co8, and Temple+, then restarted, then re-installed. Temple+ will still not launch. In fact, when I went to install Temple+, all I got was a quick flash of some notification on screen, then immediately went to the next notification telling me it was installed and the shortcuts were in my Start menu. Then, the Configuration utility came up, and I hit launch. All I get is my laptop staring at me like a cow stares at an oncoming train. It simple will not launch. Please, someone, tell me what I can do. I remember this content and would like to use it again. I have asked if numerous locations, but so far either no one has been able to help, or worse, no one has responded. Please, help.

doug1234:
Closes #873.


dolio:
So, question:

Why not just initialize the arrays directly, like here?  https://github.com/dolio/TemplePlus/tree/refs/heads/fix/feat-init

Also, `racialFeats` doesn't seem to actually be used anywhere, so just get rid of it (that's what I did there).

PR #876, opened 2025-08-07
Various Silence related fixes

dolio:
This PR includes a bunch of tweaks for the Silence spell. It's a bit old (hence the multiple base merges), so I'm going to give it a look over with regard to what's changed in the mean time before marking it as ready. Here are the current changes.

- Ported the Silence area effect function from the DLL.
    - Fixed a logic error where spell resistance was checked when _leaving_ the AoE, making you permanently silenced.
    - Only check SR for the AoE on non-strict rules
    - Don't send the 'spell end' signal when leaving the AoE. The spell doesn't actually end, and this was causing the `Dismiss` condition to be removed if the caster left the AoE.
- Added `434 - Silence.txt` to `rules/spells` with correct `Spell Resistance` and `mode_target`
- Ported the Silence begin function from the DLL
    - Made widen spell actually affect it
- Revised logic in `Spell434 - Silence.py`
    - Force spell resistance check if there is a specific target and it isn't friendly to the caster. This duplicates the check that would have happened automatically before tweaking for strict rules.
    - Also assume that a friendly target will voluntarily fail their will save, so that you can use the tactic of creating a mobile silent area
- Add a dismiss hook to the `sp-Silence` condition. It was actually impossible to dismiss in the vanilla game, I think.


doug1234:
Nice.  Is this change ready to go?

dolio:
I think so.

Though, now that I'm looking, should some of the Co8 stuff actually be in `tpgamefiles`? I've been thinking about that lately. I don't think there's anything module specific in it, but stuff it fixes won't be applied to the base game.

Issue #887, opened 2025-08-19
Spell List Not Showing Up


temyankee:
Okay. So far, my party is 7th-level. While I have not experienced that hiccup with the characters sort of leveling then not, I am still having issues with Spellcasting. There are times when I cast a spell, the character goes through the motions, but nothing happens. Additionally, when a spell has a target that I need to specify, like for Scorching Ray, the targeting indicator will be telling me the target is invalid, yet when I click, the spell fires and works as it should.
New issue: Now that my party is 7th-level, my Wizard was unable to select new Spells. There was literally no list to select from. And, my Cleric cannot select 4th-level spells to prepare, even though she has 2 available slots. There is no list of spells to select from. Now, I saved this game prior to leveling up, so I am going to try and see if it resolves itelf when I reload, which usually resolved the Spellcasting issue above as well.

Is there any file that anyone would like to see? Please let me know. Thanx.

After reloading the game, same problem. No level 4 Spells to choose from, even though my Cleric can still select a 4th-level Domain spell. Odd. Any advice/guidance would be greatly appreciated. At thi point, I am at a stand-still until I can get help with this issue.

For giggles, I went into the Co8 game directly, without Temple+, and started a party with the pregenerated characters. Used the console to level the characters, but I didn't have to get to 7th-level as I noticed something in the Co8 party that the Temple+ party did not have.

The first picture is the Co8 party. Notice how the Spell levels are listed, but not yet available as this was I believe 5th-level when I noticed this.

The second picture is the Temple+ party. Those spell levels are not there at all. Hope this helps to identify the problem.

<img width="800" height="600" alt="Image" src="https://github.com/user-attachments/assets/aff453b8-ee8b-44a8-a3c8-18d9d645abf9" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/6ded0a96-5b4c-4793-8e87-b715ad9fd538" />

Legion50:
Note: I'm not one of the devs, just a player passing through.

The invalid cursor icon still being able to target with spells is a known issue from vanilla.

I've confirmed the spell list behavior, which seems to be unique to Temple+. Empty spell level lists are hidden until the character is at the appropriate level, at which point the list is added to their spellbook.

I'm at a loss as to what's going on with your 7th-level group, but I would suggest screenshots of their spellbooks like you've provided above so that the guys here know what you're looking at.

dolio:
[...]

PR #889, opened 2025-08-22
Eliminate uses of `math.floor` in various modifier scripts.

dolio:
`math.floor` was being called on the calculated animal companion level. All this actually seems to do is convert the number to a float, because the original calculation is `ranger_level/2`, which is integer division, and automatically truncates.

Later, this value is used with `+=` on an integer-backed field from C++, which crashes with a type error (only int supported). This was causing strange errors in seemingly unrelated situations for reasons I don't completely understand.

This fixes the problem in #887. The crash seems to have been happening in the middle of the level up procedure calculating spells, resulting in no available wizard spells, and no 4th level cleric spells. I don't really understand why the Natural Bond script was being run in the middle of a _different_ character's level up, but apparently it was.

### DIFF
--- a/tpdatasrc/tpgamefiles/scr/tpModifiers/brutal_throw.py
+++ b/tpdatasrc/tpgamefiles/scr/tpModifiers/brutal_throw.py
@@ -20,10 +20,10 @@ def BrutalThrowAttackBonus(attachee, args, evt_obj):
 		
 	#Add differece between strength and dex bonus
 	strScore = attachee.stat_level_get(stat_strength)
-	strMod = math.floor((strScore - 10)/2)
+	strMod = (strScore - 10)/2
 	
 	dexScore = attachee.stat_level_get(stat_dexterity)
-	dexMod = math.floor((dexScore - 10)/2)
+	dexMod = (dexScore - 10)/2
 	
 	bonus = int(strMod - dexMod) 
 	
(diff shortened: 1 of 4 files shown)


dolio:
I took the opportunity to eliminate other uses of `math.floor` in similar situations. This means that there shouldn't be any crashes from floats getting used where the C++ end expects only integers, at least from this cause.

I verified that `(score - 10) / 2` correctly rounds down on negative numbers. So `(7 - 10) / 2 == -2`, which is the desired behavior.

temyankee:
@dolio 
So, you have in the above conversation that the Natural Bond issue was resolved. However, it still seems to be giving me issues. I have 1.0.98.0 installed. I just gave my 6th level Ranger Natural Bond on leveling up, and now that chracter's icon flashes every 6 seconds, and the Hit Point value changes by 6 then reverts itself also every 6 seconds. Double checked and if I give that same character a different Feat at level-up, and no issues. Am I missing something here? Running Temple+ and Co8. Everything else seems fine.

dolio:
@temyankee I think the answer is that this just hasn't been included in a release yet. I at least have been kind of preoccupied for a while, and that may go for Doug as well. Maybe we can pull something together in the near future.

doug1234:
@dolio Lets see about getting something out.  Its been a while now.

dolio:
Sure. I'm busy most of this week, but I'll try to get back into swing this weekend if I'm able.

temyankee:
@dolio 
Perhaps a couple of other fleshing out suggestions:
1) Maybe an option in the Temple+ Config allowing to turn off the ECL for all races.
2) I have noticed that some abiities for the suppemental/prestige classes have a note stating that the ability currently has no in game effect. Maybe finish those out? The same applies to a few feats, I believe, but I cannt think of one off the top of the noggin right off.

Also, am I using the correct methodolgy here to bring these circumstances up? I started here because that's where the Natural Bond issue had been discussed previously.

Finally, thanx again for all the groovy content and for listening/reading an old farts ramblings. Appreciate it greatly.

PR #892, opened 2025-09-17
Fix many condition related bugs

### LINKED ISSUE #886: Dead Critters Keep Making Saves Against Holy Word
Title. After all hostiles are dead and combat is ended, the combat log is still rolling with saves.

Image

dolio:
I got a little carried away on this one. It started with trying to fix #886, but while working on that I found many other things to fix.

- Paralyzed
   - Now hides the effect tool tip while Freedom of Movement is active
   - Now puts str/dex caps in place like Held. Also fixed these caps to to work correctly (before it was capping individual bonuses to 0 which still allows negatives). Also made the cap message better.
   - Now is possible to remove via Remove Paralysis. I repurposed one of the empty args to store a DC.
   - Added `HP Changed` signal on removal to recalculate things like ability score paralysis.
   - Float text when creatures are paralyzed
   - Made new applications of paralysis prefer the longer duration instead of just cancelling. Now a ghoul/ghast can repeatedly refresh the paralysis on an already paralyzed target instead of eventually allowing an escape.
- Paralyzed - Ability Score
   - Avoid adding this when the creature is already helpless, because being helpless will cap ability scores that would otherwise trigger this (rather than damage)
   - Fixed a bunch of conditions that respond positively to `Q_Helpless` even though they _don't_ make the target helpless (like, stunning conditions)
- Monster Melee Paralysis (No Elf)
   - Fixed the races tested for. They were doing some weird stuff.
   - Store DC in applied Paralyzed condition for Remove Paralysis
- Remove Paralysis
   - On strict rules, using this on more than one target only gives a new saving throw with a bonus, not guaranteed removal
   - Made Remove Paralysis work on Slow
   - Fixed a hook that would cause the target list to get messed up with multiple targets, and skip some of them
   - Made it possible to do saving throws with ad-hoc bonuses, so now you see `+4 : Remove Paralysis` in the extra saving throw, instead of a reduced DC
- Sleeping
   - Now caps some stats like Paralyzed/Held
- Deafness
   - Spell failure no longer stacks. Before if you had multiple sources of deafness you'd have to roll against each. Now there's one check based on whether or not the character is deaf.
   - Initiative penalty also no longer stacks
   - Made a non-spell version of this condition for Holy Word
- Blindness
   - Disables running/charging unless the creature has blind sight
- Summoned
   - Factored out unsummoning code for general use
   - Made summoned creatures get banished on `sp-Summoned` condition removal so they just disappear when 
successfully dispelled (instead of usually becoming hostile)
   - Made summoned creatures disappear when they die
- Holy Word
   - Made spell immediately end, with persistent conditions (this is how it's supposed to work)
   - Made deafness or being in silence protect from the non-banishment effects
   - Uses above banishment code to make creatures disappear instead of just dying.
   - Doesn't work on elemental nodes. Treating those as 'no one is on home plane' since they're pocket planes created by Zuggtmoy
- Blasphemy
   - Same general changes as Holy Word
   - Made a standalone weakness effect for this; Enfeeblement will end if there's no active spell
- Ray of Enfeeblement
   - Changed to a 'bonus' type that doesn't stack
   - Added functionality for penalty types that cannot lower the overall value below 1. These are applied _after_ other modifiers, regardless of condition order. The ray now uses this.
   - Changed the spell script to not pre-lower the penalty in light of the above
   - Can now be removed by Lesser Restoration
- Ray of Clumsiness
   - Now has a dedicated condition instead of negative Cat's Grace. Can't reduce Dex below 1.
   - Can be removed by Lesser Restoration
- Lesser Restoration
  - Now removes _all_ ability penalty conditions _or_ heals specified ability damage, preferring the former.
- Restoration
  - Now removes ability penalty conditions

Misc changes:
- You can now add conditions with object arguments in python. Objects take up two arguments.
- Allow python queries in `d20_query_with_object`
- Added `append_distinct` for (effect) tooltip events, so that you can avoid spamming if there are multiple copies of the same condition
- `args.get_param` now lets you specify a default value, with `args.get_param(i, default_val)`
- Added `is_modifier_hash` which lets you check by hash in a `EK_OnConditionAddPre` hook. Hashes can be retrieved from params, but strings can't. Use `tpdp.hash` to create them
- Added a way to put a string in the hash cache from python, `cache_name`
- Added a `cond_utils` library analogous to `spell_utils`, but for conditions that don't have an ongoing spell. Some of the above conditions are implemented this way.
- `spell_utils` tweaks to facilitate some of the above
- Deciphered the flags from `DispIoAbilityLoss` and put them in an enum

Fixes #886.
Fixes Atari bug 171 (ray of enfeeblement stacking)


dolio:
Yeah, I was going to fix up the other two Holy Word spells and then fix the merge conflict.

dolio:
Marking this back as a draft.

I realized that some of the fear/confusion stuff is probably not quite right. I was thinking that the originator of the effect could be tracked by putting the object handle in the condition. However, object handles are actually not stable across reloads. It's probably not a big deal here, because the conditions are short lived. But you can save and load in the middle of combat if you want, and it could potentially make them act weirdly. So I'm going to see if I can improve them.

PR #896, opened 2025-10-21
Avoid giving bonus equipment for more race/class proficiencies

dolio:
This modifies the Python `has_feat` implementation to allow an extra argument, specifying that only feats in the character's specific feat list should be considered. This excludes feats that the character has due to their class/race (i.e. it calls `HasFeatCount` rather than `HasFeatCountByClass`).

Using this, I modified the pc_start script to use this extra mode. This prevents new classes like beguilers and scouts from being given a weapon for each of their individual weapon proficiencies, similar to the classes with dedicated grouped proficiency feats. But, characters that specifically choose a proficiency will still get their chosen weapon. The only class this doesn't work for is (I think) Favored Soul, because their extra weapon proficiency depends on the deity, and is added with `feat_add` in a script. I don't think this is a big deal, though, because it's just a single weapon.


doug1234:
Nice fix!

